PR #6, opened 2016-03-17
delete old code

gregcaporaso:
this was confusing, we'll add back in when it works with these objects


gregcaporaso:
@jairideout or @ebolyen, can you review or merge when the tests pass? This was the source of confusion in the emails from @antgonza.

jairideout:
@gregcaporaso flake8 is failing

PR #13, opened 2016-06-22
ENH: adds a summarize visualization

gregcaporaso:
This contains figures, tables, and summary statistics that are useful for learning about what information is in a ``FeatureTable``. This was in part designed to illustrate how different types of information could be combined into a single ``Visualization``. Here's some content from the ``index.html`` file:

screenshot 2016-06-22 15 18 14


ebolyen:
Looks good! :+1:

PR #16, opened 2016-07-11
MAINT/ENH: clean-ups for alpha, adds Property('uniform_sampling')

gregcaporaso:


q2_feature_table/plugin_setup.py
```diff
@@ -19,41 +19,47 @@
...
-# be moved to the decorator calls.
 plugin.methods.register_function(
     function=q2_feature_table.rarefy,
     # TODO use more restrictive primitive type for `depth`
     inputs={'table': FeatureTable[Frequency]},
-    parameters={'depth': Int},
-    outputs=[('rarefied_table', FeatureTable[Frequency])],
+    parameters={'counts_per_sample': Int},
+    outputs=[('rarefied_table',
+              FeatureTable[Frequency] % Properties('uniform_sampling'))],
```
jairideout: What do you think about using hyphens as a naming convention for properties? `'uniform-sampling'`

Not sure if there are reasons to avoid using hyphens. @ebolyen what do you think?

q2_feature_table/plugin_setup.py
```diff
@@ -19,41 +19,47 @@
...
     name='Rarefaction',
```
jairideout: How about `name='Rarefy table'` to avoid confusion with rarefaction curves?

jairideout: Or just `name='Rarefy'`

gregcaporaso:
@jairideout, addressed those comments. I went with ``Rarefy table`` as the name - it just looked better on the cli. I can show you if you want.

PR #29, opened 2016-10-03
adds view-seq-data and view-taxa-data visualizers

gregcaporaso:


q2_feature_table/_summarize.py
```diff
@@ -14,6 +14,54 @@
...
+        fh.write('<html><body>\n')
+        # style adapted from http://stackoverflow.com/a/6667942/3424666
+        fh.write(' <style>\n'
+                 '  table {table-layout:fixed; width:1250px;}\n'
+                 '  table td {word-wrap:break-word;}\n'
+                 ' </style>\n')
+        fh.write('To BLAST a sequence against the NCBI nt database, click the '
+                 'BLAST link next to that sequence and then click the '
+                 '<i>View report</i> button on the resulting page.<p>\n')
+        fh.write(' <table border=1>\n')
```
ebolyen: 1be5d227e2011ec5cc944a839a7fb879f858e81b9e485133f89fd635c1cb9c64

q2_feature_table/_summarize.py
```diff
@@ -14,6 +14,54 @@
...
+        fh.write('  <tr><td width=300px>Feature ID</td>'
+                 '<td width=100px>BLAST link</td>'
```
ebolyen: These `td` are called `th` (for table-header) nowadays, but it doesn't really matter. There would also be a `tbody` and `thead` section, but browsers can figure it out.

gregcaporaso: Got it, thanks!

PR #38, opened 2016-10-21
removes filter method, adds filter_samples and filter_features methods

gregcaporaso:
- ~~passing a badly-formed column name seems to not work, for example: `qiime feature-table filter-samples --i-table table.qza --m-sample-metadata-file sample-metadata.tsv --p-where "\"#SampleID\"='L1S8'" --o-filtered-table junk`~~


thermokarst:
FWIW, I tested that where-clause on zsh 5.2 and bash 3.2.57(1) and the quote-escaping worked just fine:

```sh
qiime feature-table filter-samples --i-table table.qza --m-sample-metadata-file sample-metadata.tsv --p-where "\"#SampleID\"='L1S8'" --o-filtered-table junk
```

The compiled query:

```sql
SELECT "#SampleID"
FROM metadata
WHERE "#SampleID"='L1S8'
GROUP BY "#SampleID"
ORDER BY "#SampleID";
```

and the returned IDs:

```python
['L1S8']
```

q2_feature_table/_filter.py
```diff
@@ -6,16 +6,118 @@
...
+
+    conn = sqlite3.connect(':memory:')
+    conn.row_factory = lambda cursor, row: row[0]
+
+    df.to_sql('metadata', conn)
+
+    c = conn.cursor()
+    c.execute('SELECT * FROM metadata ORDER BY ROWID ASC LIMIT 1;')
+    names = [description[0] for description in c.description]
+    id_column = names[0]
```
thermokarst: Unfortunately there is no guarantee the the first column is the index column, or even that the column order will always be the same across queries. Probably makes sense to drop L24-L27 and replace with something like:

```python
id_column = df.index.name
```

q2_feature_table/_filter.py
```diff
@@ -6,16 +6,118 @@
...
+    #    us to specify complex where statements, which is what we need to do
+    #    here. For example, we need to specify things like:
+    #        WHERE Subject='subject-1' AND SampleType='gut'
+    #    but their qmark/named-style syntaxes only supports substition of
+    #    variables, such as:
+    #        WHERE Subject=?
+    # 3) sqlite3.Cursor.execute will only execute a single statement so
+    #    inserting multiple statements (e.g., "Subject='subject-1'; DROP...")
+    #    will result in an OperationalError being raised.
+    query = ('SELECT "{}" FROM metadata WHERE {} GROUP BY "{}" '
```
thermokarst: you could also use:

```python
query = ('SELECT "{0}" FROM metadata WHERE {1} GROUP BY "{0}" '
         'ORDER BY "{0}";'.format(id_column, where))
```

gregcaporaso:
> FWIW, I tested that where-clause...

@thermokarst, did you check the resulting table? The command completes for me, but I get an empty table back when it should have one sample.

gregcaporaso:
Removed my last comment - @thermokarst, the where clause does work for me too. It looks like ``qiime feature-table summarize`` barfs when there is only a single sample in the table. I was trying to use that to see if the filtering worked, and I was thinking it was barfing because there were no samples in the table. This is now #39.

q2_feature_table/_filter.py
```diff
@@ -6,16 +6,116 @@
...
+        ids_to_keep = table.ids(axis=axis)
+
+    filter_fn = _get_biom_filter_function(ids_to_keep, min_count, max_count,
+                                          min_nonzero, max_nonzero)
+    table.filter(filter_fn, axis=axis, inplace=True)
     return table
+
+
+def filter_samples(table: biom.Table, min_count: int=0, max_count: int=None,
+                   min_features: int=0, max_features: int=None,
```
gregcaporaso: `min_features` could also be called `min_nonzero`, since it's the number of nonzero values that must be in a sample vector. I think `min_features` is more clear, but I wanted to throw out the alternative of calling this `min_nonzero` (in which case `max_features` would become `max_nonzero`).

jairideout: I like `min_features`/`max_features`, it's clearer to me than the alternative for some reason.

Should `min_count`/`max_count` be `min_frequency`/`max_frequency` to match the new "frequency" nomenclature we've been using?

jairideout:
merge conflicts

q2_feature_table/_filter.py
```diff
@@ -6,16 +6,120 @@
...
+    #    inserting multiple statements (e.g., "Subject='subject-1'; DROP...")
+    #    will result in an OperationalError being raised.
+    query = ('SELECT "{0}" FROM metadata WHERE {1} GROUP BY "{0}" '
+             'ORDER BY "{0}";'.format(id_column, where))
+
+    try:
+        c.execute(query)
+    except sqlite3.OperationalError:
+        conn.close()
+        raise ValueError("Selection of ids failed with query:\n %s"
```
thermokarst: Confirm tests of this `ValueError` (e.g. non-existent column in `WHERE` clause).

q2_feature_table/_filter.py
```diff
@@ -6,16 +6,120 @@
...
+                   max_frequency: int=None, min_features: int=0,
+                   max_features: int=None,
+                   sample_metadata: qiime.Metadata=None, where: str=None)\
+                  -> biom.Table:
+    result = _filter(table=table, min_frequency=min_frequency,
+                     max_frequency=max_frequency, min_nonzero=min_features,
+                     max_nonzero=max_features, metadata=sample_metadata,
+                     where=where, axis='sample')
+    # drop features that now have zero count
+    result = _filter(table=table, min_frequency=1, max_frequency=None,
```
thermokarst: - Move this second `_filter` into itself.
- drop return val from `_filter`

q2_feature_table/_filter.py
```diff
@@ -6,16 +6,120 @@
...
+    return f
+
+
+def _filter(table, min_frequency, max_frequency, min_nonzero, max_nonzero,
+            metadata, where, axis):
+    if min_frequency == 0 and max_frequency is None and min_nonzero == 0 and\
+       max_nonzero is None and metadata is None and where is None:
+        raise ValueError("No filtering was requested.")
+    if metadata is None and where is not None:
+        raise ValueError("Metadata must be provided if where is "
```
thermokarst: denote where (`where`)

q2_feature_table/_filter.py
```diff
@@ -6,16 +6,120 @@
...
+                         "specified.")
+
+    if where is not None:
+        ids_to_keep = _ids_where(metadata, where)
+    elif metadata is not None:
+        ids_to_keep = metadata.to_dataframe().index
```
thermokarst: Add test for ids that aren't present in table.

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -122,7 +122,7 @@ def _count_summary(table, axis='sample'):
     return summary, counts
 
 
-def _get_max_count_even_sampling_depth(counts):
+def _get_max_frequency_even_sampling_depth(counts):
```
thermokarst: grep codebase for "count"

q2_feature_table/plugin_setup.py
```diff
@@ -101,6 +100,44 @@
                 "result."
 )
 
+_where_description = ("The where parameter takes a SQLite WHERE clause. "
+                      "You can get help forming these expressions at the "
+                      "following links:\n"
+                      "https://en.wikipedia.org/wiki/Where_(SQL)\n"
```
thermokarst: Link to table-filtering tutorial: `https://docs.qiime2.org/{{ qiime2.__version__ }}/tutorials/table-filtering.html`

q2_feature_table/plugin_setup.py
```diff
@@ -101,6 +100,44 @@
...
+    inputs={'table': FeatureTable[Frequency]},
+    parameters={'min_frequency': Int,
+                'max_frequency': Int,
+                'min_features': Int,
+                'max_features': Int,
+                'sample_metadata': Metadata,
+                'where': Str},
+    outputs={'filtered_table': FeatureTable[Frequency]},
+    name="Filter samples from table.",
+    description="Filter samples from table based on frequency and/or "
```
thermokarst: Provide text about default for `max_*`

jairideout:
Thanks @gregcaporaso!

PR #49, opened 2016-11-02
Use Table.sum

wasade:
Avoids cast to dense (per vector) and operates directly on the contained sparse matrix

### DIFF
--- a/q2_feature_table/plugin_setup.py
+++ b/q2_feature_table/plugin_setup.py
@@ -18,7 +18,14 @@ plugin = Plugin(
     name='feature-table',
     version=q2_feature_table.__version__,
     website='https://github.com/qiime2/q2-feature-table',
-    package='q2_feature_table'
+    package='q2_feature_table',
+    citation_text=('The Biological Observation Matrix (BIOM) format or: how '
+                   'I learned to stop worrying and love the ome-ome. '
+                   'Daniel McDonald, Jose C Clemente, Justin Kuczynski, '
+                   'Jai Ram Rideout, Jesse Stombaugh, Doug Wendel, Andreas '
+                   'Wilke, Susan Huse, John Hufnagle, Folker Meyer, Rob '
+                   'Knight and J Gregory Caporaso. GigaScience 1:7 (2012).'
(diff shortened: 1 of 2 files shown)


q2_feature_table/plugin_setup.py
```diff
@@ -18,7 +18,14 @@
     name='feature-table',
     version=q2_feature_table.__version__,
     website='https://github.com/qiime2/q2-feature-table',
-    package='q2_feature_table'
+    package='q2_feature_table',
+    citation_text=('The Biological Observation Matrix (BIOM) format or: how '
```
jairideout: I think the BIOM format citation makes more sense in q2-types (when we support multiple citations in the framework), because that is where the BIOM format transformers are implemented. This plugin provides general feature table operations that are file format-agnostic.

jairideout:
Thanks @wasade! Just one comment about the citation.

q2_feature_table/plugin_setup.py
wasade: Right, but this uses code developed for that manuscript? I added this as I had support for it from @gregcaporaso but I'm also okay moving the citation elsewhere if that's the consensus

gregcaporaso: I'm ok with it here as a lot of the operations in this plugin use the BIOM API. When we transition to having method/visualizer-specific plugins, maybe at that time we could move it from being a plugin citation to a citation for the relevant methods/visualizers.

jairideout: I still disagree with adding the citation here because q2-feature-table doesn't seem like the ideal home for it. But given that citations will work a lot differently in the future, and this citation registration will need to be updated at that point, it doesn't seem too harmful to include.

I just want to note that one of the major goals of QIIME 2 is to track relevant citations automatically for users, and to give authors credit where credit is due. I am in full support of citing the BIOM format (obviously, it needs to be cited somewhere), but it isn't clear right now where the best home for the citation is (future citation support may make this easier to determine).

wasade:
Sounds good, and agree that this isn't the ideal location -- excited to see
how citation tracking will come into effect when implemented

PR #56, opened 2016-12-09
BUG: corrections for more-strict signature checks

thermokarst:


ebolyen:
@gregcaporaso we're merging this, but it warrants further discussion later on when you are back.

PR #58, opened 2017-01-03
ENH: Add download link to raw FASTA file

### LINKED ISSUE #35: feature-table view-seq-data should allow for downloading of fasta file
A student at the workshop mentioned that he likes the BLAST links a lot, but would also like to have the fasta file in case he wants to blast some other sequence against his data. This would be easy to add.

jakereps:
Resolves #35

### DIFF
--- a/q2_feature_table/_summarize/tabulate_seqs_assets/index.html
+++ b/q2_feature_table/_summarize/tabulate_seqs_assets/index.html
@@ -8,6 +8,9 @@
         sequence and then click the <i>View report</i>
         button on the resulting page.
       <p>
+        To download a raw FASTA file of your sequences, click
+        <a href="sequences.fasta" download="sequences.fasta">here</a>.
+      </p>
       <p>
         <i>Click on a Column header to sort the table.</i>
       </p>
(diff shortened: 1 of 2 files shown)


q2_feature_table/_summarize/tabulate_seqs_assets/index.html
```diff
@@ -8,6 +8,10 @@
         sequence and then click the <i>View report</i>
         button on the resulting page.
       <p>
+      <p>
+        To download a raw FASTA file of your sequences, click
+        <a href="sequences.fasta" download="sequences.fasta">here</a>.
```
thermokarst: TIL about the [download](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download) attr!

jakereps: Same, actually. Do you know if there is a fallback for non-Chrome/Firefox/HTML5 browsers, or is this safe enough for us?

jairideout:
👍 :taco:, there's merge conflicts, ready for merge after that

jakereps:
Thanks @jairideout!

PR #60, opened 2017-01-05
Multi-commit Summarize Improvements

### LINKED ISSUE #24: summarize improvements
- [ ] add thousands separator (e.g., '{:,}'.format(1234567890), maybe think about internationalization here...)
- [ ] add frequency per feature tables
- [ ] reverse sort frequency per sample and frequency per feature tables, so highest frequencies are on top
- [ ] the width of the bars in the sample/feature count plots are confusing to users (they can't really tell that this is a histogram) when there are few samples. we should make the bin size smaller. for example:

screenshot 2016-10-12 03 07 10

- [ ] interactive exploration of rarefaction depth, as we had in q2d2. this would be really helpful to integrate with metadata, so you can determine how many samples of each metadata type a specific depth would filter from your data set
- [ ] some users at the iceland workshop noted that it would be useful to know the taxonomic composition of the samples when they're choosing their even sampling depth based on the `summarize` output - this should be possible, but would require optional artifacts and would have some overlap with the existing taxa plots (so this is lower priority)

jakereps:
- [x] add thousands separator (e.g., '{:,}'.format(1234567890), maybe think about internationalization here...)
- [x] add frequency per feature tables
- [x] reverse sort frequency per sample and frequency per feature tables, so highest frequencies are on top
- [x] the width of the bars in the sample/feature count plots are confusing to users (they can't really tell that this is a histogram) when there are few samples. we should make the bin size smaller. for example:

screenshot 2016-10-12 03 07 10

- ~interactive exploration of rarefaction depth, as we had in q2d2. this would be really helpful to integrate with metadata, so you can determine how many samples of each metadata type a specific depth would filter from your data set~
- ~some users at the iceland workshop noted that it would be useful to know the taxonomic composition of the samples when they're choosing their even sampling depth based on the `summarize` output - this should be possible, but would require optional artifacts and would have some overlap with the existing taxa plots (so this is lower priority)~

Resolves #24

### DIFF
--- a/q2_feature_table/_summarize/_visualizer.py
+++ b/q2_feature_table/_summarize/_visualizer.py
@@ -92,16 +92,24 @@ def summarize(output_dir: str, table: biom.Table) -> None:
     }
 
     sample_frequencies.sort_values(inplace=True)
+    feature_frequencies.sort_values(inplace=True)
     sample_frequencies.to_csv(
         os.path.join(output_dir, 'sample-frequency-detail.csv'))
+    feature_frequencies.to_csv(
+        os.path.join(output_dir, 'feature-frequency-detail.csv'))
 
     sample_frequencies_table = _format_html_table(
         sample_frequencies.to_frame('Frequency'))
+    feature_frequencies_table = _format_html_table(
+        feature_frequencies.to_frame('Frequency'))
     sample_frequency_template = os.path.join(
         TEMPLATES, 'summarize_assets', 'sample-frequency-detail.html')
+    feature_frequency_template = os.path.join(
+        TEMPLATES, 'summarize_assets', 'feature-frequency-detail.html')
 
-    context.update({'sample_frequencies_table': sample_frequencies_table})
-    templates = [index, sample_frequency_template]
+    context.update({'sample_frequencies_table': sample_frequencies_table,
+                    'feature_frequencies_table': feature_frequencies_table})
+    templates = [index, sample_frequency_template, feature_frequency_template]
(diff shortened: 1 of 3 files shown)


jairideout:
LGTM @jakereps! @gregcaporaso is going to test out the behavior locally and do final review/merge (probably sometime today).

gregcaporaso:
@jakereps, it looks like this hasn't been addressed: 

> reverse sort frequency per sample and frequency per feature tables, so highest frequencies are on top

I still see the lowest frequencies at the top in the csv/html versions of the *Frequency per sample detail* and *Frequency per feature detail* tables.

jakereps:
Ah! That's caused by the same misunderstanding I had when I asked if one of bullets had already been completed. I sorted the wrong tables! I'll get that changed over. This currently reverse sorts the smaller tables on the main page. Thanks @gregcaporaso!

gregcaporaso:
Yep, just noticed that it was the wrong tables. Sorry for the confusion! Could you put the summary tables back the way they were, and reverse sort the detail tables?

gregcaporaso:
Also, an even smaller bin size might be helpful in the histograms, but this is better than it was, so if theres a reason to not make the bins smaller that's fine too. 

Other than these comments, this looks good to go. I can test again when you've made the changes if you want.

jakereps:
Thanks @gregcaporaso and @jairideout! This should be ready for a final test/review.

gregcaporaso:
Looks good, thanks @jakereps!

PR #66, opened 2017-01-10
Interactive rarefaction summary

### LINKED ISSUE #63: interactive exploration of rarefaction depth in table summaries
> interactive exploration of rarefaction depth, as we had in q2d2. this would be really helpful to integrate with metadata, so you can determine how many samples of each metadata type a specific depth would filter from your data set

Taken from #24

jakereps:
~~~***WIP***~~~

Resolves #63


jairideout:
Thanks @jakereps! I think the intention was to add this interactive component to the existing `feature-table summarize` method -- @gregcaporaso is that right?

jakereps:
Yeah I saw the title get changed on the issue after I started on it. I can move it over to `summarize`. I was just throwing it up here for now to see if it was going in the right direction. Still have to implement the metadata loss.

jairideout:
Sounds good, thanks @jakereps!

gregcaporaso:
Thanks @jakereps, let me know when you'd like me to try it out and give feedback.

q2_feature_table/_summarize/summarize_assets/app/js/child.js
```diff
@@ -0,0 +1,12 @@
...
+// Copyright (c) 2016-2017, QIIME 2 development team.
+//
+// Distributed under the terms of the Modified BSD License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+// ----------------------------------------------------------------------------
+
+
+document.addEventListener('DOMContentLoaded', function() {
+  parent.postMessage(document.body.scrollHeight, '*');
```
ebolyen: This isn't quite working like it should for me, I still end up with the sub-page iframe having a scrollbar (on the bottom on normal view, and on the side when in a mobile-like resolution), instead of the visualization iframe. Might have to do with that nesting?

q2_feature_table/_summarize/summarize_assets/src/chart.js
```diff
@@ -0,0 +1,165 @@
...
+  const yAxis = axisLeft(y);
+
+  bar.append('rect')
+    .attr('fill', 'black')
+    .attr('opacity', 0.15)
+    .attr('x', d => d.x0)
+    .attr('y', d => y(d.length))
+    .attr('width', x.bandwidth() - 1)
+    .attr('height', d => props.height - y(d.length))
+  .select(function backToBar() { return this.parentNode; })
```
ebolyen: oh wow, do the callback's not provide the node as an argument? This might warrant a comment, as the correct behavior depends on how `this` is bound (requiring the old-style function)

q2_feature_table/_summarize/summarize_assets/src/chart.js
```diff
@@ -0,0 +1,165 @@
...
+    .range([props.height, 0]);
+
+  const bar = g.selectAll('.bar')
+    .data(bins)
+    .enter()
+      .append('g')
+        .attr('class', 'bar');
+        // .attr('transform', d => `translate(${d.x0}, ${y(d.length)})`);
+
+  const xAxis = axisBottom(x);
```
ebolyen: I think you may need to include this in the `updateChart` function, as the x-labels get out of sync quickly.

q2_feature_table/_summarize/summarize_assets/src/chart.js
```diff
@@ -0,0 +1,165 @@
...
+  .select(function backToBar() { return this.parentNode; })
+    .append('g')
+      .attr('class', 'overlay-group')
+    .append('rect')
+      .attr('class', 'overlay')
+      .attr('fill', 'steelblue')
+      .attr('opacity', 1)
+      .attr('stroke', 'lightgray')
+      .attr('x', d => d.x0)
+      .attr('y', d => y(d.length))
```
ebolyen: Something about this isn't working right either. It looks like the scaler is dropping stuff:
screenshot from 2017-01-23 14 22 47

q2_feature_table/_summarize/summarize_assets/src/chart.js
thermokarst: I agree with @ebolyen, but I think this whole line is unnecessary. You already have the `bar` selection from above (line 74ish). You can drop this current line (91) altogether and replace with `bar`, e.g.:

```javascript
    .attr('height', d => props.height - y(d.length));
  bar.append('g')
    etc etc etc
```

I just tested this locally and it seems to work the same as this `backToBar` callback, so personally I would just go with the already prepared selection.

thermokarst:
Playing around with this locally, I noticed that manually entering a **Sampling Depth** value into the input box doesn't force a rerender of the graph. The slider moves, but the graph stays the same (see gif below):

update

setup.py
```diff
@@ -18,6 +18,9 @@
     package_data={'q2_feature_table': ['workflows/*md'],
                   'q2_feature_table._summarize': [
                         'summarize_assets/*.html',
+                        'assets/app/*.js',
+                        'assets/app/js/*.js',
+                        'assets/app/css/*.css',
```
jakereps: Nice catch 👍  Leftover from when it was a standalone visualizer.

q2_feature_table/_summarize/summarize_assets/app/js/child.js
jakereps: [Get a real browser!](http://stackoverflow.com/questions/15932650/body-scrollheight-doesnt-work-in-firefox), haha that is really weird though, as it works as expected for me on macOS Firefox 50.1.0. I'll add those fallbacks to it and see if that helps for Ubuntu.

q2_feature_table/_summarize/summarize_assets/src/chart.js
jakereps: Not sure what you mean by this, as the x-labels don't ever change, only the y-values of each bin?

ebolyen: So when I start sampling, the values will get low enough that the coordinates of the bars will change, but the labels don't move. It also causes the old bar to "stick". So maybe it isn't this function, but the updateChart isn't handling everything correctly. `ReportedAntibioticUsage` is probably the easiest column to see this on (in the Moving Pictures data). At around 7k I see the Yes/No bars collapse into the Yes column only, leaving the No bar frozen until I bring it back. This seems to happen in other categories:

out

jakereps: Got it!

q2_feature_table/_summarize/summarize_assets/src/chart.js
jakereps: This should be fixed in the last update!

jakereps:
Thanks @ebolyen and @thermokarst! This should be ready for another look. Got it fixed by stopping trying to tweak the bars and just painting them again. Now it is 100% accurate on ones that should only have had 1 point left when dropping all but the last sample.

thermokarst:
There is still some weirdness with the bar coordinates. With the Moving Pictures dataset, category `DaysSinceExperimentStart`, the bars were overlapping for me on their initial render:

screen shot 2017-01-25 at 5 55 45 am

then as I continued to increase the threshold, one of the blue bars actually shifted over to the right, and the blue color changed (not sure its an opacity/z-index thing, or something else) to a darker shade:

screen shot 2017-01-25 at 5 55 17 am

😱

From what I can tell, the problem seems to crop up when switching from categorical data, to continuous data. I had a similar problem on the `q2-taxa barplots` viz --- I ended up treating all data as categorical data for the sake of the x-axis handling (which it looks like that's what you're doing, too. One difference I just noticed though is that the x-axis range is being set by `.rangeRound` in this viz, while in barplots I am using `.range`. I seem to recall I had some problems with `.rangeRound`, when switching between continuous and categorical data, but I could be mistaken.

jairideout:
@jakereps, looks like merging of #69 caused a conflict, sorry about that.

jakereps:
Thanks @jairideout, got it! Is there a description that you or @gregcaporaso have that I should add for the new optional `sample_metadata` parameter?

gregcaporaso:
@jakereps, we've been going with `The sample metadata.`.

jakereps:
On this one instead of individual pings, I'll go with Thanks @qiime2/core-developers! This should be ready to go with all issues addressed, including the offline catching of a bug by @ebolyen, where I found out the metadata wasn't being filtered with the feature table. Also made it resizable/responsive.

jakereps:
Ok, it was a histogram, but now it's a bar graph. Ready for a browse! 

Pair-programmed the new custom "binner" with @ebolyen.

gregcaporaso:
This is *awesome* @jakereps! I just tried this out on a real-world data set. [Here's a link](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fdl.dropboxusercontent.com%2Fu%2F2868868%2Ftemp%2Ftable-q2dev.qzv) that will temporarily work with the data set I tested on.

ebolyen:
One thing I noticed on that link, if you change categories while you have a sampling depth set, the bar-chart doesn't reflect the current value (it resets to unrarefied).

q2_feature_table/_summarize/summarize_assets/overview.html
```diff
@@ -0,0 +1,89 @@
...
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title></title>
+    <link rel="stylesheet" href="./q2templateassets/css/bootstrap.min.css">
+    <link rel="stylesheet" href="./q2templateassets/css/normalize.css">
+    <script src="./app/js/child.js" charset="utf-8"></script>
+  </head>
+  <body>
+    <div class="row">
```
ebolyen: So currently there is a tiny scrollbar on the bottom of each tab, it seems to be caused by the negative margin of `row`. I think we can remove the class (or set `margin: 0px`) for each row-div on each subpage. Just dropping the class didn't seem to have any ill-effects and did fix the bottom scrollbar. I think this is just an artifact of having an iframe in an iframe with bootstrap.

ebolyen: There are a lot of `row` nodes in each subpage, is there a top-level bootstrap container which consumes the margin correctly?

q2_feature_table/_summarize/summarize_assets/src/chart.js
```diff
@@ -0,0 +1,154 @@
...
+
+const categoricalBin = (data, counts, x) => {
+  const result = [];
+  x.domain().forEach((cat) => {
+    const catArray = Object.keys(data)
+                      .filter(sampleID => data[sampleID] === cat)
+                      .map(sampleID => counts[sampleID]);
+    catArray.x0 = x(cat);
+    result.push(catArray);
+  });
```
ebolyen: :+1: much nicer than yesterday!

q2_feature_table/_summarize/summarize_assets/src/chart.js
```diff
@@ -0,0 +1,154 @@
...
+    .style('fill-opacity', 1)
+    .attr('stroke', 'lightgray')
+    .attr('x', d => d.x0)
+    .attr('y', d => y(d.filter(count => count >= threshold).length))
+    .attr('width', x.bandwidth() - 1)
+    .attr('height', d => props.height - y(d.filter(count => count >= threshold).length));
+
+  const hiddenDepths = d3.select('#hidden-depths');
+  hiddenDepths
+    .on('change', () => updateChart(bar, props, y, +hiddenDepths.node().value));
```
ebolyen: I think we need to updateChart at the end, so that if the category changes (causing a rebuild) the existing depth is used (it hasn't changed, so the on-change handler won't fire)

q2_feature_table/_summarize/summarize_assets/sample-frequency-detail.html
```diff
@@ -1,9 +1,46 @@
...
     </div>
-  </div>
-{% endblock %}
+    <div class="row">
+      <div id="table" class="col-lg-10 col-md-10 col-sm-12 col-xs-12 pull-right"></div>
+    </div>
+    <script type="text/javascript">
+      var m;
+      var c;
+      function load(metadata, counts) {
```
ebolyen: I think you can simplify this if you just changed the jsonp to call `init` and then rename `init.js` to `main.js`. 
Then _I think_ you can drop this entire script block.

q2_feature_table/_summarize/summarize_assets/webpack.config.js
```diff
@@ -0,0 +1,34 @@
...
+
+module.exports = {
+  entry: {
+    app: './src/main.js',
+    vendor: ['d3', 'javascript-natural-sort']
+  },
+  plugins: [
+    new webpack.optimize.CommonsChunkPlugin('vendor', 'app/vendor.bundle.js'),
+    new webpack.optimize.UglifyJsPlugin({
+      compress: { warnings: false }
```
ebolyen: adding:
```javascript
mangle: {
    except: ['init']
}
```
to the uglifyjs config should allow you to use the JSONP directly instead of using single-letter globals to avoid the mangle (if I understand what's happening correctly).

q2_feature_table/_summarize/summarize_assets/sample-frequency-detail.html
ebolyen: Also, we'd need an exception to the mangle-config of uglifyjs (comment below)

ebolyen:
@jakereps, continued an existing review, so you have to scroll up slightly. I don't know why github didn't just put it at the bottom...

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -46,12 +46,22 @@ def tabulate_seqs(output_dir: str, data: DNAIterator) -> None:
     shutil.copy(js, os.path.join(output_dir, 'js', 'tsorter.min.js'))
 
 
-def summarize(output_dir: str, table: biom.Table) -> None:
-    number_of_samples = table.shape[1]
-    number_of_features = table.shape[0]
+def summarize(output_dir: str, table: pd.DataFrame,
```
jakereps: Got it swapped and reused some things that were already calculated to build the jsonp. I think was because I originally built this as a new visualizer, and ported the code over. Thought it'd take more to redo what I had in `biom` than to convert everything else to `pd`, but after just swapping it back it wasn't, as everything was already being computed further down the line. On a quick test `biom` is faster on the Moving Pictures table. 
image

q2_feature_table/_summarize/summarize_assets/overview.html
jakereps: That's it, the `q2templates` has the container, and this was the weird hacked together combination of it to be a child, which forgot the container div. Should that `q2templates` child frame we had talked about happen before merging this?

q2_feature_table/_summarize/summarize_assets/src/chart.js
jakereps: Now that I look at it, is it weird to be using this `x.domain().forEach()` method, instead of a `for (const cat of x.domain()) {}`? Same thing, mainly just readability maybe?

q2_feature_table/_summarize/summarize_assets/src/chart.js
jakereps: Oops, that was there, but got dropped in the refactor. Will put it back!

q2_feature_table/_summarize/summarize_assets/sample-frequency-detail.html
jakereps: I can't get this to play nice. Searching the `bundle.js` there is a non-mangled `init` method, but the jsonp (modified from `load(` to `init(` and placing the `bundle.js` above the jsonp loading) won't call it.

ebolyen: Very strange, the script execution should be sequential...

q2_feature_table/_summarize/summarize_assets/src/chart.js
ebolyen: I'm good with either.

q2_feature_table/_summarize/summarize_assets/overview.html
ebolyen: Whichever is faster I think. I wouldn't block this PR over q2templates though.

q2_feature_table/_summarize/summarize_assets/sample-frequency-detail.html
jakereps: The bundle itself is wrapped, so as-is nothing inside is callable from the page itself. 

Found something that might solve it though (http://stackoverflow.com/a/34361312). I'll try it out after class.

ebolyen: oh is it possible that `init` isn't global because of how webpack handles imports? What if you did `window['init'] = function(){...}`

q2_feature_table/_summarize/summarize_assets/overview.html
jakereps: Missed this response before my last push. Fixed this one for now!

jakereps:
Thanks @ebolyen! Everything has been addressed, and few bugs (`df`/`biom` shape output order)/fixes found.

gregcaporaso:
Awesome, thanks guys! Our users are going to love this!

ebolyen:
It is so cool! :tada:

PR #71, opened 2017-01-31
BUG: Bin number type issue with seaborn or numpy

jakereps:
As of numpy 1.12.0 using floats with `np.zeros` has been removed, which caused our bins value to break `sns.distplot`, because `round` returned a float.


jairideout:
👍 , thanks @jakereps!

PR #90, opened 2017-02-27
TST: add .coveragerc

gregcaporaso:


gregcaporaso:
Coverage goes down here because, prior to the addition of `.coveragerc`, coverage was being computed with the `test_*.py` files included, which were always 100% coverage.

PR #98, opened 2017-04-19
BUG: fixed sequences retained in summarize

### LINKED ISSUE #97: error in Sequences retained calculation
Sequences retained should be number of samples retained times even sampling depth, but it's not:

screenshot 2017-04-19 08 28 20

ebolyen:
Fixes #97 

image

image

image


gregcaporaso:
The results look correct to me. Someone else should review the code.

PR #100, opened 2017-05-03
ENH: Clarify error message when table is rarefied to nothing

### LINKED ISSUE #87: Error message resulting from choosing too high a sampling depth is not clear.
image
The error message that is obtained when using a value for sampling depth that is to high is vague and doesn't really point out what the problem is. I found it by looking at the sampling depth qzv and csv. It was giving this 2-d 3-d error when I chose a sampling depth that I thought was reasonable, but it turns out wasn't.
Thanks,
Arron

jakereps:
Addresses part of #87 

image


jakereps:
Thanks @jairideout! I added a test to this!

PR #102, opened 2017-05-08
ENH: Add qualitative counts to feature summary

### LINKED ISSUE #85: viz: see how many samples a feature is observed in
This came up on the forum [here](https://forum.qiime2.org/t/feature-request-sequence-per-sample-question/339). One way to accomplish this is to compute qualitative, as opposed to quantitative, stats in `feature-table summarize`. This is similar to `biom summarize-table --qualitative --observations`. Instead of a flag we can just compute qualitative and quantitative stats in the viz.

jakereps:
image

Resolves #85

### DIFF
--- a/q2_feature_table/_summarize/_visualizer.py
+++ b/q2_feature_table/_summarize/_visualizer.py
@@ -113,7 +113,7 @@ def summarize(output_dir: str, table: biom.Table,
 
     feature_frequencies = feature_frequencies.astype(int) \
         .apply('{:,}'.format).to_frame('Frequency')
-    feature_frequencies['Qualitative Observations'] = \
+    feature_frequencies['# of Samples Observed In'] = \
         pd.Series(feature_qualitative_data).astype(int).apply('{:,}'.format)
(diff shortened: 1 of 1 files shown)


gregcaporaso:
I think we should change _Qualitative Observations_ to something that will make more sense to our users. Maybe _Number of samples_ or _Samples observed in_ or ... (something like that, I'm not happy with either of those).

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -147,6 +151,14 @@ def summarize(output_dir: str, table: biom.Table,
         fh.write(');')
 
 
+def _compute_qualitative_summary(table):
+    table = table.transpose()
+    sample_count = {}
+    for count_vector, sample_id, metadata in table.iter():
+        sample_count[sample_id] = (count_vector != 0).sum()
```
gregcaporaso: Should this be `sample_count[feature_id]` (I think it's just a variable name change).

jakereps:
Thanks @gregcaporaso, I've addressed your comments and tried a different verbiage on the column title.

PR #105, opened 2017-05-17
ENH: Change sample_metadata to metadata

ElDeveloper:
All other plugins that I checked refer to this as "metadata" not
"sample_metadata".

--------

~~Would like to let the tests run in Travis.~~

### DIFF
--- a/q2_feature_table/plugin_setup.py
+++ b/q2_feature_table/plugin_setup.py
@@ -216,7 +216,7 @@ plugin.methods.register_function(
                 'max_frequency': Int,
                 'min_samples': Int,
                 'max_samples': Int,
-                'feature_metadata': Metadata,
+                'metadata': Metadata,
                 'where': Str},
     outputs=[('filtered_table', FeatureTable[Frequency])],
     input_descriptions={
@@ -235,8 +235,8 @@ plugin.methods.register_function(
                         'be observed in to be retained. If no value is '
                         'provided this will default to infinity (i.e., no '
                         'maximum sample filter will be applied).'),
-        'feature_metadata': 'Feature metadata used in conjuction with `where` '
-                            'parameter to select features to retain.',
+        'metadata': 'Feature metadata used in conjuction with `where` '
+                    'parameter to select features to retain.',
         'where': 'SQLite WHERE clause specifying feature metadata criteria '
                  'that must be met to be included in the filtered feature '
                  'table. If not provided, all features in `feature_metadata` '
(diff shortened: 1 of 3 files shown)


jairideout:
This naming is intentional because `filter-samples` applies only to the "sample" axis, and `filter-features` applies to the "feature" axis (`filter-features` has a corresponding `feature_metadata` option).

We plan to implement more general filtering in the future (at the framework level) so this interface will change, but for now we need to have axis-specific filtering methods.

ElDeveloper:
Got it!

jairideout:
Thinking about this more, I'm in favor of merging these changes. @ElDeveloper you're right that we're using `metadata` instead of axis-specific names in other methods/visualizers, and the axis the metadata applies to is implied by the filtering method name (i.e. `filter-samples` implies the metadata will apply to samples, and likewise with `filter-features`). The parameter documentation also reflects which axis the metadata applies to.

Can you please make the same changes to `filter-features`? Then this is GTG. Thanks @ElDeveloper!

q2_feature_table/plugin_setup.py
```diff
@@ -192,8 +192,8 @@
                          'have to be retained. If no value is provided '
                          'this will default to infinity (i.e., no maximum '
                          'feature filter will be applied).'),
-        'sample_metadata': 'Sample metadata used in conjuction with `where` '
-                           'parameter to select samples to retain.',
+        'metadata': 'Sample metadata used in conjuction with `where` parameter'
+                    ' to select samples to retain.',
         'where': 'SQLite WHERE clause specifying sample metadata criteria '
                  'that must be met to be included in the filtered feature '
                  'table. If not provided, all samples in `sample_metadata` '
```
jairideout: Can you change `` `sample_metadata` `` to `` `metadata` `` on this line?

q2_feature_table/plugin_setup.py
```diff
@@ -235,8 +235,8 @@
                         'be observed in to be retained. If no value is '
                         'provided this will default to infinity (i.e., no '
                         'maximum sample filter will be applied).'),
-        'feature_metadata': 'Feature metadata used in conjuction with `where` '
-                            'parameter to select features to retain.',
+        'metadata': 'Feature metadata used in conjuction with `where` '
+                    'parameter to select features to retain.',
         'where': 'SQLite WHERE clause specifying feature metadata criteria '
                  'that must be met to be included in the filtered feature '
                  'table. If not provided, all features in `feature_metadata` '
```
jairideout: Can you change `` `feature_metadata` `` to `` `metadata` `` on this line?

jairideout:
Thanks @ElDeveloper! 🎉 👍

PR #112, opened 2017-06-08
ENH: Optional flag for exclusion filters, Fixes #101

### LINKED ISSUE #101: filter-samples/filter-features: support "negate" filtering option
QIIME 1's filtering scripts have an option to negate the filtering operation (i.e. *exclude* instead of *include*). It'd be useful to add that functionality to `filter-samples`/`filter-features`.

This came up on the forum [here](https://forum.qiime2.org/t/removing-deleting-a-single-feature-from-a-feature-table/560).

maxvonhippel:
This PR addresses Issue #101 .

I've added an optional `exclude_ids` flag to `filter_features` and `filter_samples`.  I declared the new flag in the plugin setup and added unit tests.

This flag is set to `False` by default.  If it is specified as true, then the filter function in question returns the inverse of what it would otherwise return.  For example:

````PYTHON
df = pd.DataFrame({'SequencedGenome': ['yes']},
                          index=['O1'])
metadata = qiime2.Metadata(df)
table = Table(np.array([[0, 1, 3], [1, 1, 2]]),
                      ['O1', 'O2'],
                      ['S1', 'S2', 'S3'])
filter_features(table, metadata=metadata)
````

... returns:

````BASH
#OTU ID	S2	S3
O1	1.0	3.0
````
This makes sense: we asked for the stuff in `index=['01']`, and we got it.

In comparison, if we set `exclude_ids=True`, we get:

````PYTHON
filter_features(table, metadata=metadata, exclude_ids=True)
````

````BASH
#OTU ID	S1	S2	S3
O2	1.0	1.0	2.0
````

And this also makes sense: we asked for the stuff *not* in `index=['01']`, and, indeed, we got everything not in that index (in this case the only other index being `02`).

### DIFF
--- a/q2_feature_table/_filter.py
+++ b/q2_feature_table/_filter.py
@@ -13,7 +13,6 @@ import numpy as np
 
 def _get_biom_filter_function(ids_to_keep, min_frequency, max_frequency,
                               min_nonzero, max_nonzero):
-    ids_to_keep = set(ids_to_keep)
     if max_frequency is None:
         max_frequency = np.inf
     if max_nonzero is None:
@@ -30,18 +29,25 @@ _other_axis_map = {'sample': 'observation', 'observation': 'sample'}
 
 
 def _filter(table, min_frequency, max_frequency, min_nonzero, max_nonzero,
-            metadata, where, axis):
+            metadata, where, axis, exclude_ids=False):
     if min_frequency == 0 and max_frequency is None and min_nonzero == 0 and\
        max_nonzero is None and metadata is None and where is None:
         raise ValueError("No filtering was requested.")
     if metadata is None and where is not None:
         raise ValueError("Metadata must be provided if 'where' is "
                          "specified.")
+    if metadata is None and exclude_ids is True:
+        raise ValueError("Metadata must be provided if 'exclude_ids' "
+                         "is specified")
 
     if metadata is not None:
         ids_to_keep = metadata.ids(where=where)
     else:
         ids_to_keep = table.ids(axis=axis)
+    if exclude_ids is True:
+        ids_to_keep = set(table.ids(axis=axis)) - set(ids_to_keep)
+    else:
+        ids_to_keep = set(ids_to_keep)
 
     filter_fn1 = _get_biom_filter_function(
         ids_to_keep, min_frequency, max_frequency, min_nonzero, max_nonzero)
@@ -58,12 +64,13 @@ def _filter(table, min_frequency, max_frequency, min_nonzero, max_nonzero,
 def filter_samples(table: biom.Table, min_frequency: int=0,
                    max_frequency: int=None, min_features: int=0,
                    max_features: int=None,
-                   metadata: qiime2.Metadata=None, where: str=None)\
+                   metadata: qiime2.Metadata=None, where: str=None,
+                   exclude_ids: bool=False)\
                   -> biom.Table:
     _filter(table=table, min_frequency=min_frequency,
             max_frequency=max_frequency, min_nonzero=min_features,
             max_nonzero=max_features, metadata=metadata,
-            where=where, axis='sample')
+            where=where, axis='sample', exclude_ids=exclude_ids)
 
     return table
 
@@ -71,11 +78,12 @@ def filter_samples(table: biom.Table, min_frequency: int=0,
 def filter_features(table: biom.Table, min_frequency: int=0,
                     max_frequency: int=None, min_samples: int=0,
                     max_samples: int=None,
-                    metadata: qiime2.Metadata=None, where: str=None)\
+                    metadata: qiime2.Metadata=None, where: str=None,
+                    exclude_ids: bool=False)\
                    -> biom.Table:
     _filter(table=table, min_frequency=min_frequency,
             max_frequency=max_frequency, min_nonzero=min_samples,
             max_nonzero=max_samples, metadata=metadata,
-            where=where, axis='observation')
+            where=where, axis='observation', exclude_ids=exclude_ids)
 
     return table
--- a/q2_feature_table/_filter.py
+++ b/q2_feature_table/_filter.py
@@ -41,10 +41,6 @@ def _filter(table, min_frequency, max_frequency, min_nonzero, max_nonzero,
     if metadata is None and exclude_ids is True:
         raise ValueError("Metadata must be provided if 'exclude_ids' "
                          "is true.")
-    if where is not None and exclude_ids is True:
-        raise ValueError("'exclude_ids' must be false if 'where' is "
-                         "specified.")
-
     if metadata is not None:
         ids_to_keep = metadata.ids(where=where)
     else:
--- a/q2_feature_table/_filter.py
+++ b/q2_feature_table/_filter.py
@@ -13,6 +13,7 @@ import numpy as np
 
 def _get_biom_filter_function(ids_to_keep, min_frequency, max_frequency,
                               min_nonzero, max_nonzero):
+    ids_to_keep = set(ids_to_keep)
     if max_frequency is None:
         max_frequency = np.inf
     if max_nonzero is None:
@@ -31,14 +32,18 @@ _other_axis_map = {'sample': 'observation', 'observation': 'sample'}
 def _filter(table, min_frequency, max_frequency, min_nonzero, max_nonzero,
             metadata, where, axis, exclude_ids=False):
     if min_frequency == 0 and max_frequency is None and min_nonzero == 0 and\
-       max_nonzero is None and metadata is None and where is None:
+       max_nonzero is None and metadata is None and where is None and\
+       exclude_ids is False:
         raise ValueError("No filtering was requested.")
     if metadata is None and where is not None:
         raise ValueError("Metadata must be provided if 'where' is "
                          "specified.")
     if metadata is None and exclude_ids is True:
         raise ValueError("Metadata must be provided if 'exclude_ids' "
-                         "is specified")
+                         "is true.")
+    if where is not None and exclude_ids is True:
+        raise ValueError("'exclude_ids' must be false if 'where' is "
+                         "specified.")
 
     if metadata is not None:
         ids_to_keep = metadata.ids(where=where)
@@ -46,8 +51,6 @@ def _filter(table, min_frequency, max_frequency, min_nonzero, max_nonzero,
         ids_to_keep = table.ids(axis=axis)
     if exclude_ids is True:
         ids_to_keep = set(table.ids(axis=axis)) - set(ids_to_keep)
-    else:
-        ids_to_keep = set(ids_to_keep)
 
     filter_fn1 = _get_biom_filter_function(
         ids_to_keep, min_frequency, max_frequency, min_nonzero, max_nonzero)
--- a/q2_feature_table/_filter.py
+++ b/q2_feature_table/_filter.py
@@ -40,7 +40,7 @@ def _filter(table, min_frequency, max_frequency, min_nonzero, max_nonzero,
                          "specified.")
     if metadata is None and exclude_ids is True:
         raise ValueError("Metadata must be provided if 'exclude_ids' "
-                         "is true.")
+                         "is True.")
     if metadata is not None:
         ids_to_keep = metadata.ids(where=where)
     else:
--- a/q2_feature_table/plugin_setup.py
+++ b/q2_feature_table/plugin_setup.py
@@ -200,10 +200,9 @@ plugin.methods.register_function(
                  'that must be met to be included in the filtered feature '
                  'table. If not provided, all samples in `metadata` that are '
                  'also in the feature table will be retained.',
-        'exclude_ids': 'If `True`, returns the filtered set of samples '
-                       'which occur in the feature table but not in the '
-                       'metadata (or the filtered metadata, if invoked '
-                       'in conjunction with `where`).'
+        'exclude_ids': 'If `True`, the samples selected by `metadata` or '
+                       '`where` parameters will be excluded from the filtered '
+                       'table instead of being retained.'
     },
     output_descriptions={
         'filtered_table': 'The resulting feature table filtered by sample.'
@@ -249,10 +248,9 @@ plugin.methods.register_function(
                  'that must be met to be included in the filtered feature '
                  'table. If not provided, all features in `metadata` that are '
                  'also in the feature table will be retained.',
-        'exclude_ids': 'If `True`, returns the filtered set of features '
-                       'which occur in the feature table but not in the '
-                       'metadata (or the filtered metadata, if invoked '
-                       'in conjunction with `where`).'
+        'exclude_ids': 'If `True`, the features selected by `metadata` or '
+                       '`where` parameters will be excluded from the filtered '
+                       'table instead of being retained.'
     },
     output_descriptions={
         'filtered_table': 'The resulting feature table filtered by feature.'
--- a/q2_feature_table/plugin_setup.py
+++ b/q2_feature_table/plugin_setup.py
@@ -200,11 +200,10 @@ plugin.methods.register_function(
                  'that must be met to be included in the filtered feature '
                  'table. If not provided, all samples in `metadata` that are '
                  'also in the feature table will be retained.',
-        'exclude_ids': 'If `True`, function returns the filtered set of '
-                       'samples which occur in the feature table but not in '
-                       'the metadata (or not in the metadata which satisfies '
-                       'the `where` parameter, if it is provided). `False` by '
-                       'default.'
+        'exclude_ids': 'If `True`, returns the filtered set of samples '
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: q2_feature_table/_filter.py [hunk 1]
```diff
@@ -13,7 +13,6 @@
 
 def _get_biom_filter_function(ids_to_keep, min_frequency, max_frequency,
                               min_nonzero, max_nonzero):
-    ids_to_keep = set(ids_to_keep)
```
maxvonhippel: By the time we get here, `ids_to_keep` have become a set no matter what, so this is redundant.

q2_feature_table/_filter.py
```diff
@@ -30,18 +29,25 @@ def f(data_vector, id_, metadata):
...
+                         "is specified")
 
     if metadata is not None:
         ids_to_keep = metadata.ids(where=where)
     else:
         ids_to_keep = table.ids(axis=axis)
+    if exclude_ids is True:
+        ids_to_keep = set(table.ids(axis=axis)) - set(ids_to_keep)
+    else:
+        ids_to_keep = set(ids_to_keep)
```
maxvonhippel: If `exclude_ids` is True, we create a set.  If not, we need to turn `its_to_keep` into a set, since it might not be already.  This operation was previously in `_get_biom_filter_function`, but I think it makes sense here, since it highlights the fact that it's already a set in the case where we are doing the subtraction operation because `exclude_ids` is True.  If you disagree let me know; I'm happy to change it back.

q2_feature_table/tests/test_filter.py
```diff
@@ -203,6 +203,44 @@ def test_sample_metadata(self):
         expected = Table(np.array([]), [], [])
         self.assertEqual(actual, expected)
 
+        # exclude one
```
maxvonhippel: All of the below tests are based on those for filtering (e.g, `exclude one` is based on `filter one`, etc), the difference being that I expect the opposite result since `exclude_ids=True`.

jairideout:
Just a quick note that coverage decreased a little -- @maxvonhippel can you please look into that?

thermokarst:
Thanks for the reminder about code coverage, @jairideout! @maxvonhippel, I think I will have a few additional tests for you to add --- stay tuned for a review.

q2_feature_table/tests/test_filter.py
```diff
@@ -276,6 +314,11 @@ def test_where(self):
         expected = Table(np.array([]), [], [])
         self.assertEqual(actual, expected)
 
+        # exclude all and no metadata
+        with self.assertRaises(ValueError):
```
thermokarst: Can you make this an `assertRaisesRegex` --- I don't think you are catching the `Exception` that you think you are catching (hint: check out the test coverage decrease reported by coveralls). You might want to move this test into a different test group too.

q2_feature_table/tests/test_filter.py
```diff
@@ -456,6 +525,11 @@ def test_where(self):
...
```
thermokarst: Same comment as above: can you make this an `assertRaisesRegex`

q2_feature_table/tests/test_filter.py
thermokarst: Test cases added in this PR so far:
- Sample axis:
  - exclude one
  - exclude two
  - exclude all
  - incorrect invocation (this test needs to be revised, see comments below)
- Feature axis:
  - exclude one
  - exclude two
  - exclude all
  - incorrect invocation (this test needs to be revised, see comments below)

Can you add the following tests:
- Sample axis:
  - exclusion with min frequency (confirm behavior with samples above an below threshold)
  - exclusion with max frequency (confirm behavior with samples above an below threshold)
  - exclusion with min and max frequency (confirm behavior with samples above an below threshold)
  - exclusion with min features (confirm behavior with samples above an below threshold)
  - exclusion with max features (confirm behavior with samples above an below threshold)
  - exclusion with min and max features (confirm behavior with samples above an below threshold)
  - exclusion with `where` clause

I think we can get away with the tests that you have already written for the Feature axis, for the same reasons as mentioned in [the comment](https://github.com/qiime2/q2-feature-table/pull/112/files#diff-4ad12f722dbdc707bd5cb97fea327c93R388).

Let me know if you need help setting these tests up.

q2_feature_table/_filter.py
```diff
@@ -30,18 +29,25 @@ def f(data_vector, id_, metadata):
...
-            metadata, where, axis):
+            metadata, where, axis, exclude_ids=False):
     if min_frequency == 0 and max_frequency is None and min_nonzero == 0 and\
        max_nonzero is None and metadata is None and where is None:
         raise ValueError("No filtering was requested.")
     if metadata is None and where is not None:
         raise ValueError("Metadata must be provided if 'where' is "
                          "specified.")
+    if metadata is None and exclude_ids is True:
+        raise ValueError("Metadata must be provided if 'exclude_ids' "
```
thermokarst: 👍

q2_feature_table/_filter.py
thermokarst: Sounds good to me! You could also drop the `else` and always cast (one line less):

```python
if exclude_ids is True:
    ids_to_keep = set(table.ids(axis=axis)) - set(ids_to_keep)
ids_to_keep = set(ids_to_keep)
```

If you went that route, you could also keep the `ids_to_keep = set(ids_to_keep)` in `_get_biom_filter_function` (less changes in the `git diff`).

Your call!

q2_feature_table/plugin_setup.py
```diff
@@ -197,7 +198,10 @@
         'where': 'SQLite WHERE clause specifying sample metadata criteria '
                  'that must be met to be included in the filtered feature '
                  'table. If not provided, all samples in `metadata` that are '
-                 'also in the feature table will be retained.'
+                 'also in the feature table will be retained.',
+        'exclude_ids': 'If `True`, function returns the filtered set of '
```
maxvonhippel: I like your articulation of that @thermokarst .

maxvonhippel:
Ready for review! @jairideout

q2_feature_table/tests/test_filter.py
maxvonhippel: Thanks for the comprehensive list!  And done :)

maxvonhippel:
I was mistaken on Friday about the order of operations with `exclude_ids` + `where`.  The code allowed the combination of these parameters already, and I had unnecessarily hard-coded in an exception for a reasonable, well-defined, and already working scenario.  I removed that exception and did the following:

1. Re-arranged the tests for `exclude_ids` to make more sense.  Individual testing sets now end with the case where `exclude_ids` is included in the test, for example, 

````PYTHON
def test_invalid_args(self):
        table = Table(np.array([[0, 1, 3], [1, 1, 2]]),
                      ['O1', 'O2'],
                      ['S1', 'S2', 'S3'])
        # various tests
        with self.assertRaisesRegex(ValueError,
                                    "Metadata must be provided if "
                                    "'exclude_ids' is true."):
            filter_samples(table, exclude_ids=True)
````
... and there is an entire section in the `FilterSamplesTests` devoted to combinations of sample filters and `exclude_ids=True`:
````PYTHON
def test_combine_id_and_frequency_filters(self):
        # no filtering
        df = pd.DataFrame({'Subject': ['subject-1', 'subject-1', 'subject-2'],
                           'SampleType': ['gut', 'tongue', 'gut']},
                          index=pd.Index(['S1', 'S2', 'S3'], name='#SampleID'))
        metadata = qiime2.Metadata(df)
        table = Table(np.array([[0, 1, 3], [1, 1, 2]]),
                      ['O1', 'O2'],
                      ['S1', 'S2', 'S3'])
        where = "Subject='subject-1' OR Subject='subject-2'"
        actual = filter_samples(table, metadata=metadata, where=where,
                                min_frequency=1)
        expected = Table(np.array([[0, 1, 3], [1, 1, 2]]),
                         ['O1', 'O2'],
                         ['S1', 'S2', 'S3'])
        self.assertEqual(actual, expected)
# etc etc
````
2. I have improved the comments which denote what each test does.  In the case where I first apply some filter with `where`, and then invert that filter with `exclude_ids=True`, I denote the test like in this example:
````PYTHON
# where filter one -> exclude one,
# max_features filter one,
# min_features filter one
````
The arrow here is meant to show that filtering happens first and then is inverted in order to exclude items.  If that syntax is unclear please let me know and I will try to figure out a better way to express the idea.

3. I have arranged the tests for `exclude_ids=True` to follow a fairly consistent pattern:
````
metadata exclude one, min_frequency exclude none
metadata exclude one, min_frequency exclude one
metadata exclude one, max_frequency exclude none
metadata exclude one, max_frequency exclude one
metadata exclude one, min_frequency exclude none, max_frequency exclude none
metadata exclude one, min_frequency exclude one, max_frequency exclude one
where filter one -> exclude one, min_frequency exclude one, max_frequency exclude one
# and then repeat this pattern for min_features and max_features
````

4. I improved the `test_invalid_args` to test not just for a `ValueError` in each case but for the specific `ValueError` that is expected.

File: q2_feature_table/tests/test_filter.py [hunk 8]
```diff
@@ -340,6 +437,252 @@ def test_combine_frequency_filters(self):
...
+        where = "Subject='subject-1'"
+        actual = filter_samples(table,
+                                metadata=metadata,
+                                exclude_ids=True,
+                                where=where,
+                                max_frequency=4,
+                                min_frequency=3)
+        expected = Table(np.array([]), [], [])
+        self.assertEqual(actual, expected)
+
```
thermokarst: can you split these into two?

```python
def test_combine_exclude_ids_and_frequency_filters):
    foo
    ...

def test_combine_exclude_ids_and_features_filters):
    foo
    ....
```

thermokarst: Also, a few more tests in this section:

- frequency filtering:
  - exclude one by id, filter same sample by min filter (variation of exclude one, min_frequency filter one)
  - exclude one by id, filter sample sample by max filter
- feature filtering:
  - exclude one by id, filter same sample by min filter (variation of exclude one, min_frequency filter one)
  - exclude one by id, filter sample sample by max filter

q2_feature_table/tests/test_filter.py
```diff
@@ -276,6 +322,57 @@ def test_where(self):
         expected = Table(np.array([]), [], [])
         self.assertEqual(actual, expected)
 
+        # filter none -> exclude none
```
thermokarst: Can you add one more test case in this section: filter all -> exclude all

maxvonhippel: Yep, can do!

q2_feature_table/tests/test_filter.py
maxvonhippel: To be clear, you want me to demonstrate that if I exclude id `S3` and then pass a min filter that would only let S3 through, I should end up with nothin, since I excluded S3 and nothing else satisfies that filter, right?  That's what you'd like me to demonstrate, f or min and then max, that you can't basically "filter back in" something you just excluded by id?

maxvonhippel: Example:
````PYTHON
# this test already exists
# exclude one, min_frequency filter one
df = pd.DataFrame({'Subject': ['subject-1'],
                   'SampleType': ['gut']},
                  index=['S1'])
metadata = qiime2.Metadata(df)
table = Table(np.array([[0, 1, 3], [1, 1, 2]]),
              ['O1', 'O2'],
              ['S1', 'S2', 'S3'])
actual = filter_samples(table,
                        metadata=metadata,
                        exclude_ids=True,
                        min_frequency=3)
expected = Table(np.array([[3], [2]]),
                 ['O1', 'O2'],
                 ['S3'])
self.assertEqual(actual, expected)

# this test is what I believe you are suggesting
# exclude one, min_frequency filter for same one
df = pd.DataFrame({'Subject': ['subject-1'],
                  'SampleType': ['gut']},
                  index=['S1'])
metadata = qiime2.Metadata(df)
table = Table(np.array([[10, 1, 3], [1, 1, 2]]),
              ['O1', 'O2'],
              ['S1', 'S2', 'S3'])
actual = filter_samples(table,
                        metadata=metadata,
                        exclude_ids=True,
                        min_frequency=10)
expected = Table(np.array([]), [], [])
self.assertEqual(actual, expected)
````

**Comment by maxvonhippel** [comment]
@thermokarst Ok, I just implemented the changes you requested and pushed.  Ready for review.  Thanks!

File: q2_feature_table/_filter.py [hunk 10]
```diff
@@ -30,18 +30,23 @@ def f(data_vector, id_, metadata):
...
+       max_nonzero is None and metadata is None and where is None and\
+       exclude_ids is False:
         raise ValueError("No filtering was requested.")
     if metadata is None and where is not None:
         raise ValueError("Metadata must be provided if 'where' is "
                          "specified.")
-
+    if metadata is None and exclude_ids is True:
+        raise ValueError("Metadata must be provided if 'exclude_ids' "
+                         "is true.")
```
thermokarst: `s/true/True/g`

q2_feature_table/plugin_setup.py
```diff
@@ -235,12 +243,18 @@
                         'be observed in to be retained. If no value is '
                         'provided this will default to infinity (i.e., no '
                         'maximum sample filter will be applied).'),
-        'metadata': 'Feature metadata used in conjuction with `where` '
-                    'parameter to select features to retain.',
+        'metadata': 'Sample metadata used with `where` parameter when '
```
thermokarst: `filter_features` filters by feature metadata, not sample metadata --- can you please update this text to match that?

q2_feature_table/tests/test_filter.py
thermokarst: Sorry @maxvonhippel, I missed another one here: exclude none. This would looks like metadata with index values that aren't found in the table.

q2_feature_table/_filter.py
maxvonhippel: Would you like it to say

> if 'exclude_ids' is True.

... or ...

> if 'exclude_ids' is 'True'.

?

q2_feature_table/plugin_setup.py
maxvonhippel:
````
'metadata': 'Feature metadata used with `where` parameter when '
                    'selecting features to retain, or with `exclude_ids` '
                    'when selecting features to discard.',
````

Good?

File: q2_feature_table/_filter.py (hunk 10 above)
Comment (maxvonhippel) [inline q2_feature_table/_filter.py]: (My thought would be the former, `True`, but just want to double check)

Comment (thermokarst) [inline q2_feature_table/_filter.py]: Sorry, I assumed you were familiar with that substitution syntax (that means globally substitute the word `True` for `true`). So, your first option presented, please!

**Comment by maxvonhippel** [comment]
Ok @thermokarst I updated the code accordingly and pushed, thanks!

File: q2_feature_table/plugin_setup.py (hunk 7 above)
Comment (thermokarst) [inline q2_feature_table/plugin_setup.py]: Sorry, I peeked through some of the other plugins/actions out there in the ecosystem, maybe something like this:

> If `True`, returns the filtered set of samples which occur in the feature table but not in the metadata (or the filtered metadata, if invoked in conjunction with `where`).

Note, we don't need to say anything about the default value --- that is for the interfaces to handle.

File: q2_feature_table/plugin_setup.py [hunk 12]
```diff
@@ -235,12 +243,18 @@
...
-                    'parameter to select features to retain.',
+        'metadata': 'Feature metadata used with `where` parameter when '
+                    'selecting features to retain, or with `exclude_ids` '
+                    'when selecting features to discard.',
         'where': 'SQLite WHERE clause specifying feature metadata criteria '
                  'that must be met to be included in the filtered feature '
                  'table. If not provided, all features in `metadata` that are '
-                 'also in the feature table will be retained.'
+                 'also in the feature table will be retained.',
+        'exclude_ids': 'If `True`, function returns the filtered set of '
```
thermokarst: > If `True`, returns the filtered set of features which occur in the feature table but not in the metadata (or the filtered metadata, if invoked in conjunction with `where`).

q2_feature_table/plugin_setup.py
```diff
@@ -235,12 +242,17 @@
...
-                    'parameter to select features to retain.',
+        'metadata': 'Feature metadata used with `where` parameter when '
+                    'selecting features to retain, or with `exclude_ids` '
+                    'when selecting features to discard.',
         'where': 'SQLite WHERE clause specifying feature metadata criteria '
                  'that must be met to be included in the filtered feature '
                  'table. If not provided, all features in `metadata` that are '
-                 'also in the feature table will be retained.'
+                 'also in the feature table will be retained.',
+        'exclude_ids': 'If `True`, returns the filtered set of features '
```
maxvonhippel: I prefer your wording @jairideout .

thermokarst: Sounds good to me!

PR #113, opened 2017-06-13
MAINT: clarified meaning of zero sampling depth

### LINKED ISSUE #110: Interactive Sample Detail, question/bug?
Ported from https://github.com/qiime2/qiime2/issues/276 (@antgonza)

During the ASM workshop a user pointed out that the "Sequences Retained: 0 (0.00%)" field is confusing, which I agree. The confusing points are:

- It starts in 0 with 0.0% retained, where it should be 100%, right?
- Once you start moving to the right, it seems to be going up so showing the sequences removed (not retained) and then once you get to around (50%) it starts going up, see screenshot.

feat-table

gregcaporaso:
Fixes #110 

A few screenshots:

screenshot 2017-06-13 10 37 14

screenshot 2017-06-13 10 37 06


q2_feature_table/_summarize/summarize_assets/src/controls.js
```diff
@@ -81,8 +91,7 @@ const initializeControls = () => {
     .attr('display', 'inline-block')
     .style('padding-top', '10px')
```
ebolyen: If you made the above simplification, you should be able to do `.append('span')` right after the `.style` call.

ebolyen: Then instead of the rest, you would just call `updateState(trows, slider)`

q2_feature_table/_summarize/summarize_assets/src/controls.js
```diff
@@ -41,8 +52,7 @@ const initializeControls = () => {
 
   const updateStats = (trows, slider) => {
     formGroup.selectAll('span')
-      .data([['Samples Retained: ', `${calcSampleRetainment(trows, slider).join(' (')}%)`],
-             ['Sequences Retained: ', `${calcFeatureRetainment(trows, slider).join(' (')}%)`]])
+      .data([['Retained ', `${calcFeatureRetainment(trows, slider).join(' (')}%)`, ' sequences in ', `${calcSampleRetainment(trows, slider).join(' (')}%)`, ' samples at the specifed sampling depth.']])
```
ebolyen: I think all of this can be simplified to:
```javascript
const sequences = `${calcFeatureRetainment(trows, slider).join(' (')}%)`
const samples = `${calcSampleRetainment(trows, slider).join(' (')}%)`
fromGroup.selectAll('span')
  .text(`Retained ${sequences} sequences in ${samples} samples at the specifed sampling depth.`)
  .insert('br')

ebolyen:
My review seems to have skipped the submission of some of the comments...

q2_feature_table/_summarize/summarize_assets/src/controls.js
ebolyen: This was *supposed* to be in the review, not sure what happened.

q2_feature_table/_summarize/summarize_assets/src/controls.js
ebolyen: (above simplification is actually below because reasons...)

PR #127, opened 2017-08-18
ENH: add `core-features` visualizer

jairideout:
Example .qzv: [core-features.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fforum-qiime2-org.s3-us-west-2.amazonaws.com%2Foriginal%2F1X%2F89efca24e973aa4b164cddc9ce31360ddea942de.qzv)

Ported @gregcaporaso's code from https://github.com/qiime2/q2-diversity/pull/127 and made the following changes:

- All references to "core microbiome" are now "core features".

- Removed metadata-based filtering support from the viz in favor of using `filter-samples` or `filter-features`.

- Refactored unit tests to reduce code duplication, and added several new tests.

- Fixed a precision bug where small enough step sizes could cause "feature list" .tsv output files to be overwritten.

- Fixed a bug where HTML table contents could have been truncated.

- Better handling of the case where `min_fraction` and `max_fraction` are the same value (silenced a matplotlib warning, and ignores `steps` for efficiency and easier to interpret output).

- Updated code to use biom API more efficiently.

Fixes https://github.com/qiime2/q2-diversity/issues/122.

### DIFF
--- a/q2_feature_table/_core_features/_visualizer.py
+++ b/q2_feature_table/_core_features/_visualizer.py
@@ -98,10 +98,6 @@ def _get_core_features(table, fraction):
 
 
 def _get_filter_to_core_f(table, fraction):
-    if not (0.0 <= fraction <= 1.0):
-        raise ValueError("Invalid fraction passed to core filter: %r is "
-                         "outside of range [0,1]." % fraction)
-
     # determine the number of samples that must have a non-zero value for a
     # feature to be considered part of the core
     min_count = fraction * table.shape[1]
--- a/q2_feature_table/_core_features/_visualizer.py
+++ b/q2_feature_table/_core_features/_visualizer.py
@@ -1,0 +1,171 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016-2017, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import os.path
+import pkg_resources
+
+import biom
+import q2templates
+import numpy as np
+import pandas as pd
+import seaborn as sns
+
+
+TEMPLATES = pkg_resources.resource_filename(
+    'q2_feature_table._core_features', 'core_features_assets')
+
+
+def core_features(output_dir, table: biom.Table, min_fraction: float=0.5,
+                  max_fraction: float=1.0, steps: int=11) -> None:
+    if max_fraction < min_fraction:
+        raise ValueError('min_fraction (%r) parameter must be less than '
+                         'max_fraction (%r) parameter.' %
+                         (min_fraction, max_fraction))
+
+    index_fp = os.path.join(TEMPLATES, 'index.html')
+    context = {
+        'num_samples': table.shape[1],
+        'num_features': table.shape[0]
+    }
+
+    if min_fraction == max_fraction:
+        fractions = [min_fraction]
+    else:
+        fractions = np.linspace(min_fraction, max_fraction, steps)
+
+    rounded_fractions = _round_fractions(fractions)
+
+    data = []
+    file_links = []
+    for fraction, rounded_fraction in zip(fractions, rounded_fractions):
+        core_features = _get_core_features(table, fraction)
+        core_feature_count = len(core_features)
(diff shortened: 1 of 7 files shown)


q2_feature_table/_core_features/core_features_assets/index.html
```diff
@@ -0,0 +1,40 @@
...
+    </div>
+  </div>
+  <div class='col-md-6'>
+    <div class='row'>
+      <div class='col-md-12'>
+        <h3>Terminology key</h3>
+        <i>Fraction of samples</i>: The fraction of the total number of samples that a feature must be observed more than zero times in for that feature to be considered "core".<br>
+        <i>Feature count</i>: The number of "core" features at a given fraction of samples.<br>
+        <i>Fraction of features</i>: The fraction of the total number of features that are "core" features at a given fraction of samples.<br>
+        <i>Feature list</i>: The list of "core" feature ids and their percentile frequencies at a given fraction of samples.
```
ebolyen: `frequencies at a given fraction of samples` -> `frequencies across a given fraction of samples`?

ebolyen: Actually maybe all of these `at`s would read better as `across`?

q2_feature_table/_core_features/_visualizer.py
```diff
@@ -0,0 +1,171 @@
...
+        return pd.DataFrame(data, index=index).sort_values(
+            by='50%', ascending=False)
+    else:
+        return pd.DataFrame()
+
+
+def _get_filter_to_core_f(table, fraction):
+    if not (0.0 <= fraction <= 1.0):
+        raise ValueError("Invalid fraction passed to core filter: %r is "
+                         "outside of range [0,1]." % fraction)
```
jairideout: > Assuming min_fraction and max_fraction are in [0, 1] could np.linspace ever produce values that would hit this branch?
>
> The range predicates on the QIIME type will enforce the value on the min/max arguments, but if we wanted to check anyways, it seems like this logic should just be at the top of the method.

Good catch! This validation is unnecessary so I'll remove it.

> Anecdotally, the compositional statistics book "Modeling and Analysis of Compositional Data" calls these kinds of values proportions very consistently. (Not important for this PR, but it's the closest I've seen as a name for the [0, 1] range.)

Discussed offline with @ebolyen and @thermokarst. It's kind of a toss-up between using *fraction of X* vs *proportion of X* so we'll keep using *fraction* for consistency with other QIIME 2 plugins (and QIIME 1 scripts). We can revisit this later if there's need to switch terms.

q2_feature_table/_core_features/core_features_assets/index.html
jairideout: I think the current wording makes sense -- I would also be open to changing it to `frequencies for a given fraction of samples`, both wordings seem equally clear to me. Using `across` doesn't make sense because each term (i.e. column label) is referring to core features at a single fraction of samples -- IOW, each row represents core features at a single fraction of samples (i.e. a single *step*). `across` seems to imply that the core features span fractions of samples. Does that make sense?

jairideout: Discussed offline with @ebolyen and we decided that `across` doesn't make sense, and to drop the phrase "at a given fraction of samples" since it could be confused with computing the percentile frequencies using only those samples where the "core" feature is observed. We'll also add the word "total" to make this clearer.

q2_feature_table/_core_features/core_features_assets/index.html
```diff
@@ -0,0 +1,40 @@
...
+        Feature count: {{num_features}}
+      </div>
+    </div>
+    <div class='row'>
+      <div class='col-md-12'>
+        <img src='core-feature-counts.svg'>
+      </div>
+    </div>
+  </div>
+  <div class='col-md-6'>
```
jairideout: Discussed offline with @ebolyen, we're leaving the page layout as-is because the plot is intended to be the first thing users see, and the table provides a more detailed view of what's in the plot. The reflow for small screens also works well as-is because the plot is placed at the top of the page.

jairideout:
@ebolyen thanks for the review! I addressed your comments in a separate "squash" commit and updated the example .qzv linked in the PR description.

Issue #120, opened 2017-08-04
BUG: Can't download FASTA file from tabulate-seqs


thermokarst:
The convenience link for downloading the source FASTA [appears to be broken](https://forum.qiime2.org/t/cant-download-fasta-file-from-feature-table-tabulate-seqs/955?u=thermokarst) (at least when used on q2view).

thermokarst:
Update --- this is specifically an issue with q2view and this visualization --- viewing locally via q2cli works as expected.

jairideout:
Fixed in #130 :tada:

PR #130, opened 2017-08-22
BUG: replace download attr for tabulate-seqs

ebolyen:
Works in `qiime tools view` (causes the download dialog to fire). 

q2view opens a new tab, but we can probably teach it what a `.fasta` file is to avoid that

### DIFF
--- a/q2_feature_table/_summarize/tabulate_seqs_assets/index.html
+++ b/q2_feature_table/_summarize/tabulate_seqs_assets/index.html
@@ -9,7 +9,7 @@
         button on the resulting page.
       <p>
         To download a raw FASTA file of your sequences, click
-        <a href="sequences.fasta" download="sequences.fasta">here</a>.
+        <a href="sequences.fasta" target="_blank">here</a>.
       </p>
       <p>
         <i>Click on a Column header to sort the table.</i>
--- a/q2_feature_table/_summarize/tabulate_seqs_assets/index.html
+++ b/q2_feature_table/_summarize/tabulate_seqs_assets/index.html
@@ -9,7 +9,7 @@
         button on the resulting page.
       <p>
         To download a raw FASTA file of your sequences, click
-        <a href="sequences.fasta" target="_blank">here</a>.
+        <a href="sequences.fasta" target="_blank" rel="noopener noreferrer">here</a>.
       </p>
       <p>
         <i>Click on a Column header to sort the table.</i>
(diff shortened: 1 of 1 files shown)


ebolyen:
fixes #120

q2_feature_table/_summarize/tabulate_seqs_assets/index.html
```diff
@@ -9,7 +9,7 @@
         button on the resulting page.
       <p>
         To download a raw FASTA file of your sequences, click
-        <a href="sequences.fasta" download="sequences.fasta">here</a>.
+        <a href="sequences.fasta" target="_blank">here</a>.
```
jakereps: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

> Note: When using target, consider adding rel="noopener noreferrer" to avoid exploitation of the window.opener API.

ebolyen: I'm don't this really applies here as we're not opening a document (which means no `window.opener` api to exploit), it's a fasta file. Happy to add it though.

jakereps: Oops, submitted with no description. Someone could make a shady fasta file 🕵️ , might want to include that `rel="noopener noreferrer"` to any `target="_blank"` attrs across all plugins.

https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

jakereps: I personally have no preference. Just something I stumbled across way back that I've personally made sure to include any time I use a `target=_blank`, so just threw it out there. ¯\\\_(ツ)\_/¯

q2_feature_table/_summarize/tabulate_seqs_assets/index.html
```diff
@@ -25,7 +25,7 @@
         {% for sequence in data %}
           <tr>
             <td>{{ sequence.id }}</td>
-            <td><a target="_blank" href="{{ sequence.url }}">{{ sequence.seq }}</a></td>
+            <td><a target="_blank" href="{{ sequence.url }}" rel="noopener noreferrer">{{ sequence.seq }}</a></td>
```
ebolyen: thanks @jakereps, I noticed this one, which definitely needs the noopener!

PR #135, opened 2017-08-28
ENH: adds group method

### LINKED ISSUE #86: port `collapse_samples.py` functionality from QIIME 1
This can be used for collapsing replicate groups, or for combining samples that are sequenced in different sequencing runs/lanes.

ebolyen:
This allows you to group a feature-table using a metadata category which
maps existing ids into new bins. This can be done on either axis of the
table.

fixes #86

### DIFF
--- a/q2_feature_table/_group.py
+++ b/q2_feature_table/_group.py
@@ -56,6 +56,9 @@ def _munge_metadata_category(mc, ids):
 
 def group(table: biom.Table, axis: str, metadata: qiime2.MetadataCategory,
           mode: str) -> biom.Table:
+    if table.is_empty():
+        raise ValueError("Cannot group an empty table.")
+
     if axis == 'feature':
         axis = 'observation'
 
--- a/q2_feature_table/_group.py
+++ b/q2_feature_table/_group.py
@@ -26,7 +26,7 @@ _mode_lookup = {
 }
 
 
-def _munge_metadata_category(mc, ids):
+def _munge_metadata_category(mc, ids, axis):
     # TODO: centralize these ideas in MetadataCategory somehow
     series = mc.to_series()
     table_ids = set(ids)
@@ -39,8 +39,9 @@ def _munge_metadata_category(mc, ids):
     # Check missing
     missing_ids = table_ids - series_ids
     if missing_ids:
-        raise ValueError("All IDs must be present in metadata category, but "
-                         " the following are missing: %r" % missing_ids)
+        raise ValueError("All %s IDs in the feature table must be present in "
+                         "the metadata category, but the following are "
+                         "missing: %r" % (axis, missing_ids))
 
     # While preserving order, get rid of any IDs found only in the metadata
     series = series.drop(series_ids - table_ids)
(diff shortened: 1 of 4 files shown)


q2_feature_table/_group.py
```diff
@@ -0,0 +1,71 @@
...
+def _collapse_factory(function):
+    def collapse_f(table, axis):
+        # axis is always the transpose of the original collapse axis
+        return np.array([function(x) for x in table.iter_data(axis=axis)])
+    return collapse_f
+
+
+_mode_lookup = {
+    'sum': _collapse_factory(np.sum),
+    'median-ceiling': _collapse_factory(lambda x: np.ceil(np.median(x))),
```
ebolyen: I'm a little torn here. It feel like it would make more sense for the median to pick the "right-hand" value in the event of an even number of ids. Then all of the observation frequencies in the collapsed table are strictly values that were found in the original. (Why introduce a mean of two numbers only to take the ceiling of it? It just invents a number out of thin air...)

Apparently any value between and include the two middle values is an adequate definition of median as it minimizes the distance between all other values: https://math.stackexchange.com/a/47806. So while the above implementation isn't wrong, it seems like there's a simpler and more reasonable definition of median (pick the right hand number) that maintains closure over `FeatureTable[Frequency]`.

q2_feature_table/plugin_setup.py
```diff
@@ -94,6 +94,37 @@
...
+        'axis': Str % Choices({'sample', 'feature'})
+    },
+    outputs=[
+        ('grouped_table', FeatureTable[Frequency])
+    ],
+    input_descriptions={
+        'table': 'The table to group samples or features on.'
+    },
+    parameter_descriptions={
+        'mode': 'How to combine observation frequencies within group.',
```
ebolyen: This is probably confusing terminology, is there a better word for the cells of a contingency table? Also, I should describe the meaning of the modes, now that they aren't as clear-cut

ebolyen: Googling around I found "joint frequency" which kind of makes sense.

q2_feature_table/_group.py
jairideout: Did your discussion with @gregcaporaso resolve this?

ebolyen: No this came afterwards, just wanted to bring this up, what we have isn't wrong so we can run with it.

q2_feature_table/_group.py
```diff
@@ -0,0 +1,79 @@
...
+    'mean-ceiling': _collapse_factory(lambda x: np.ceil(np.mean(x)))
+}
+
+
+def _munge_metadata_category(mc, ids, axis):
```
thermokarst: :+1:

q2_feature_table/tests/test_group.py
```diff
@@ -0,0 +1,402 @@
...
+
+import biom
+import qiime2
+import pandas as pd
+import numpy as np
+
+from q2_feature_table import group
+
+
+class TestGroup(unittest.TestCase):
```
thermokarst: ❤️ these tests!

ebolyen:
Changes look great, thanks @jairideout!

PR #140, opened 2017-09-13
ENH: Support filtering FeatureData[Sequence]

### LINKED ISSUE #119: support filtering FeatureData[Sequence]
Feature request came up [on the forum](https://forum.qiime2.org/t/filtering-representative-sequences/825). Add a method to support filtering sequences from `FeatureData[Sequence]` artifacts. This is something we'll need for feature parity with QIIME 1.

thermokarst:
Fixes #119 

Note to reviewer: I packaged and split up the filtering tests, since this suite was getting large (and in charge).


q2_feature_table/_filter.py
```diff
@@ -29,8 +30,8 @@ def f(data_vector, id_, metadata):
 _other_axis_map = {'sample': 'observation', 'observation': 'sample'}
 
 
-def _filter(table, min_frequency, max_frequency, min_nonzero, max_nonzero,
-            metadata, where, axis, exclude_ids=False):
+def _filter_table(table, min_frequency, max_frequency, min_nonzero,
```
thermokarst: I renamed this util to disambiguate from the other types of filtering this module now performs.

q2_feature_table/_filter.py
```diff
@@ -80,9 +81,19 @@ def filter_features(table: biom.Table, min_frequency: int=0,
...
-            where=where, axis='observation', exclude_ids=exclude_ids)
+    _filter_table(table=table, min_frequency=min_frequency,
+                  max_frequency=max_frequency, min_nonzero=min_samples,
+                  max_nonzero=max_samples, metadata=metadata,
+                  where=where, axis='observation', exclude_ids=exclude_ids)
 
     return table
+
+
+def filter_seqs(data: pd.Series, metadata: qiime2.Metadata, where: str=None,
```
thermokarst: As implemented, this feature performs id-based filtering (with or without a `where` clause), so it made sense to me to make `metadata` required.

q2_feature_table/plugin_setup.py
```diff
@@ -300,6 +300,41 @@
                 "on https://docs.qiime2.org for additional details."
 )
 
+plugin.methods.register_function(
+    function=q2_feature_table.filter_seqs,
+    inputs={
+        'data': FeatureData[Sequence],
```
thermokarst: I'm not in love with `data`, but we use this in a [few](https://github.com/qiime2/q2-feature-table/blob/master/q2_feature_table/plugin_setup.py#L158-L159) places in [this](https://github.com/qiime2/q2-feature-table/blob/master/q2_feature_table/plugin_setup.py#L316) plugin.

q2_feature_table/_filter.py
jairideout: Looks good, this matches how `qiime diversity filter-distance-matrix` works (it only performs ID- and metadata-based filtering).

q2_feature_table/plugin_setup.py
jairideout: I'm not in love either -- what name would you choose if there wasn't a precedent for `data`?

q2_feature_table/_filter.py
ebolyen: Should this method be called `filter_seq_data` to match the merge operations on non-table objects in this plugin?

q2_feature_table/_filter.py
```diff
@@ -80,9 +81,19 @@ def filter_features(table: biom.Table, min_frequency: int=0,
...
+                exclude_ids: bool=False) -> pd.Series:
+    # Note, no need to check for missing feature IDs in the metadata, because
+    # that is basically the point of this method.
+    ids_to_keep = metadata.ids(where=where)
+    if exclude_ids is True:
+        ids_to_keep = set(data.index) - set(ids_to_keep)
+    return data[data.index.isin(ids_to_keep)]
```
ebolyen: What happens when the result is completely empty (perhaps they supplied their sample metadata instead of feature ids). It seems like we may want to error if none of the IDs matched.

q2_feature_table/plugin_setup.py
```diff
@@ -300,6 +300,41 @@
...
+    },
+    parameter_descriptions={
+        'metadata': 'Feature metadata used for id-based filtering, with '
+                    '`where` parameter when selecting features to retain, or '
+                    'with `exclude_ids` when selecting features to discard.',
+        'where': 'SQLite WHERE clause specifying feature metadata criteria '
+                 'that must be met to be included in the filtered feature '
+                 'table. If not provided, all features in `metadata` that are '
+                 'also in the sequences will be retained.',
+        'exclude_ids': 'If `True`, the features selected by `metadata` or '
```
ebolyen: `True` depends on the interface, I would go with just "true"

q2_feature_table/tests/filter/test_filter_sequences.py
```diff
@@ -0,0 +1,102 @@
...
+from pandas.util.testing import assert_series_equal
+
+from q2_feature_table import filter_seqs
+
+
+class FilterSeqsTests(unittest.TestCase):
+    def setUp(self):
+        self.seqs = pd.Series(['ACGT', 'GCTA', 'CCCC', 'TGTT'],
+                              index=['O1', 'O2', 'O3', 'O4'])
+        self.md_lite = pd.DataFrame(
```
ebolyen: `self.df_lite` instead? Since you have to make it metadata first

q2_feature_table/tests/filter/test_filter_sequences.py
```diff
@@ -0,0 +1,102 @@
...
+            [['A'], ['C'], ['G'], ['T']], index=['O1', 'O2', 'O3', 'O4'],
+            columns=['seq'])
+        md_full = pd.DataFrame(
+            [['foo', '1'], ['bar', '2'], ['baz', '3'], ['foo', '4']],
+            index=['O1', 'O2', 'O3', 'O4'], columns=['stuff', 'some_numbers'])
+        md_full.index.name = 'FeatureID'
+        self.md_full = qiime2.Metadata(md_full)
```
ebolyen: It would be good to see a metadata object where none of the IDs matched at all.

q2_feature_table/tests/filter/test_filter_sequences.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def assertFiltered(self, exp, md=None, exclude_ids=False,
```
ebolyen: This name is kind of confusing to me, I really expect it to perform only a comparison (of some kind), not run the function that we're testing.

What if it was 
```python
def expect_after_filter(...):
```

or something like that?

Alternatively it looks like there's a fair bit of wiring to save 2 lines in practice, maybe we don't need this?

q2_feature_table/_filter.py
thermokarst: @ebolyen, i'm fine with that, @jairideout, what do you think?

q2_feature_table/_filter.py
thermokarst: No error right now: it looks like our feature table filtering doesn't error when empty (I could be wrong there, though), but distance matrix filtering does error. I am fine either way.

q2_feature_table/plugin_setup.py
thermokarst: 👍 This was copied from the other `filter` methods in this plugin. Want me to make this change in all three locations?

q2_feature_table/tests/filter/test_filter_sequences.py
thermokarst: I originally wrote the tests without this helper, in my opinion it was pretty clunky to try and read the tests and understand what was going one. I agree about the wiring, but I felt like the readability payoff was worth it.

q2_feature_table/tests/filter/test_filter_sequences.py
thermokarst: There are a couple of examples of that in the tests ([with](https://github.com/qiime2/q2-feature-table/pull/140/files#diff-179288ed1a08aed39629f8e221d51606R82) and [without](https://github.com/qiime2/q2-feature-table/pull/140/files#diff-179288ed1a08aed39629f8e221d51606R48) `where`) - maybe I am misunderstanding the request, though.

q2_feature_table/tests/filter/test_filter_sequences.py
thermokarst: What about `filter_and_assertEqual`?

q2_feature_table/plugin_setup.py
thermokarst: I like `seqs`, but then it becomes a bit stuttery with the action name...

Maybe `data` isn't the worst thing ever.

q2_feature_table/_filter.py
jairideout: I like `filter-seqs` because it matches `tabulate-seqs` and is shorter than `filter-seq-data`. The `-data` suffix doesn't really add much here, except for promoting arthritis!

I'd be in favor of dropping `-data` from `merge-seq-data` and `merge-taxa-data`, though that's outside the scope of this PR (may be a good change to make soon if others are on board with it, since it's user-facing). If we went with this strategy, the new method names could be `merge-seqs` and `merge-taxa`, respectively.

q2_feature_table/plugin_setup.py
jairideout: Makes sense, let's just keep `data` (relates to [my comment](https://github.com/qiime2/q2-feature-table/pull/140#discussion_r138739903) above).

q2_feature_table/_filter.py
ebolyen: I like that plan!

q2_feature_table/_filter.py
jairideout: I'd prefer to error (for now, at least) until https://github.com/qiime2/q2-types/issues/144 is resolved. `filter-features`/`filter-samples` should also get updated at some point, though outside the scope of this PR.

ebolyen: I'm kind of thinking maybe the feature-table filters should error as well. There's nothing you can do with an empty feature table anyways, so it's just a delayed error. That's out of scope for this PR of course.

q2_feature_table/plugin_setup.py
ebolyen: That would be awesome!

q2_feature_table/tests/filter/test_filter_sequences.py
ebolyen: I was thinking more along the lines of the metadata has nothing in common with the artifact (vs an empty filter from where). I get that they'll hit the same code path, but it seems like the most common way to mess this up.

thermokarst: Isn't that the test linked above under [`without`](https://github.com/qiime2/q2-feature-table/pull/140/files#diff-179288ed1a08aed39629f8e221d51606R48)? That metadata has nothing in common with the artifact.

ebolyen: I apologize, I only looked at the `with`. My bad!

q2_feature_table/_filter.py
jairideout: Jinx!

thermokarst:
Thanks @ebolyen (and @jairideout)! I have addressed your comments (included as a SQUASH commit). Please let me know if I missed anything, or if you have any additional changes you would like to see. Thanks!!!

PR #142, opened 2017-09-28
ENH: Support merging on overlapping axes (sum)

### LINKED ISSUE #137: `merge`: support merging on overlapping sample and feature axes
Current handling of overlapping axes is limited to sample overlap or feature overlap, but not both.

thermokarst:
Fixes #137

### DIFF
--- a/q2_feature_table/tests/test_merge.py
+++ b/q2_feature_table/tests/test_merge.py
@@ -93,6 +93,52 @@ class MergeTableTests(unittest.TestCase):
         with self.assertRaisesRegex(ValueError, 'overlap method'):
             merge(t1, t2, 'peanut')
 
+    def test_sum_full_overlap(self):
+        t1 = Table(np.array([[0, 1, 3], [1, 1, 2]]),
+                   ['O1', 'O2'],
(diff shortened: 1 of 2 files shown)


q2_feature_table/tests/test_merge.py
```diff
@@ -93,6 +93,52 @@ def test_invalid_overlap_method(self):
         with self.assertRaisesRegex(ValueError, 'overlap method'):
             merge(t1, t2, 'peanut')
 
+    def test_sum_full_overlap(self):
```
gregcaporaso: Could you add a `test_sum_partial_overlap`, where some but not all of the sample ids _and_ some but not all of the features ids overlap? I don't expect any issues, but I think that's a very common use case in practice, so it makes sense to have a test for it.

thermokarst: Good catch @gregcaporaso - addressed in a [squash commit](https://github.com/qiime2/q2-feature-table/pull/142/commits/8684501f26c59c24076f150f581b10981cc6a927) 🎃 .

gregcaporaso: 👍

PR #143, opened 2017-09-28
Fixes #129

### LINKED ISSUE #129: `feature-table rarefy`: use `Range` for `sampling_depth` parameter
`feature-table rarefy` currently defines `sampling_depth` as `Int`. Update to use `Int % Range(1, None)` so that the type system can detect and disallow sampling depths less than 1.

wasade:
Sets an integer `Range` on sampling depth. I was not sure if and how to add a unit test for this as I believe this is controlled at plugin method registration. Below is the output I receive now:

```bash
(qiime2-2017.8) $ qiime feature-table rarefy --i-table table.qza --o-rarefied-table test --p-sampling-depth 5
Saved FeatureTable[Frequency] to: test.qza
(qiime2-2017.8.redux) $ qiime feature-table rarefy --i-table table.qza --o-rarefied-table test --p-sampling-depth 0
Usage: qiime feature-table rarefy [OPTIONS]

Error: Invalid value for "--p-sampling-depth": 0 is smaller than the minimum valid value 1.
(qiime2-2017.8.redux) $ qiime feature-table rarefy --i-table table.qza --o-rarefied-table test --p-sampling-depth -1
Usage: qiime feature-table rarefy [OPTIONS]

Error: Invalid value for "--p-sampling-depth": -1 is smaller than the minimum valid value 1.
```

### DIFF
--- a/q2_feature_table/plugin_setup.py
+++ b/q2_feature_table/plugin_setup.py
@@ -35,7 +35,7 @@ plugin = Plugin(
 plugin.methods.register_function(
     function=q2_feature_table.rarefy,
     inputs={'table': FeatureTable[Frequency]},
-    parameters={'sampling_depth': Int},
+    parameters={'sampling_depth': Int % Range(1, None)},
     outputs=[('rarefied_table', FeatureTable[Frequency])],
     input_descriptions={'table': 'The feature table to be rarefied.'},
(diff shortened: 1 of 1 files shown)


wasade:
Awesome!

Issue #144, opened 2017-09-28
Bump to BIOM 2.1.6


wasade:
QIIME2 currently depends on BIOM 2.1.5. Bumping to 2.1.6 (see [changelog](https://github.com/biocore/biom-format/blob/master/ChangeLog.md#biom-216) would bring in a few useful additions such as `Table.rankdata`, `Table.concat`, and `Table.del_metadata` while also resolving a few nagging bugs in the HDF5 handling.

jairideout:
Done! https://github.com/qiime2/environment-files/tree/master/latest/staging

PR #146, opened 2017-09-28
Support for randomly sampling features or samples from a table

wasade:
This adds in support for randomly subsampling a table by the IDs. One particularly useful scenario is within a meta-analysis when you need to balance group sizes between tables.

### DIFF
--- a/q2_feature_table/_sample.py
+++ b/q2_feature_table/_sample.py
@@ -26,4 +26,3 @@ def sample(table: biom.Table, sampling_depth: int, axis: str) -> biom.Table:
         table = table.transpose()
 
     return table
-
--- a/q2_feature_table/_sample.py
+++ b/q2_feature_table/_sample.py
@@ -1,0 +1,29 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016-2017, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import biom
+
+
+def sample(table: biom.Table, sampling_depth: int, axis: str) -> biom.Table:
+    if axis == 'feature':
+        # we are transposing the table due to biocore/biom-format#759
+        table = table.transpose()
+
+    # the axis is always 'sample' due to the above transpose
+    table = table.subsample(sampling_depth, axis='sample', by_id=True)
+
+    # the inverted axis is always observation due to the above transpose
+    invaxis = 'observation'
+    table.filter(lambda v, i, m: v.sum() > 0, axis=invaxis)
+
+    if axis == 'feature':
+        # reverse the transpose necessary due to biocore/biom-format#759
+        table = table.transpose()
+
+    return table
+
--- a/q2_feature_table/_subsample.py
+++ b/q2_feature_table/_subsample.py
@@ -9,13 +9,14 @@
 import biom
 
 
-def sample(table: biom.Table, sampling_depth: int, axis: str) -> biom.Table:
+def subsample(table: biom.Table, subsampling_depth: int,
+              axis: str) -> biom.Table:
     if axis == 'feature':
         # we are transposing the table due to biocore/biom-format#759
         table = table.transpose()
 
     # the axis is always 'sample' due to the above transpose
-    table = table.subsample(sampling_depth, axis='sample', by_id=True)
+    table = table.subsample(subsampling_depth, axis='sample', by_id=True)
 
     # the inverted axis is always observation due to the above transpose
     invaxis = 'observation'
--- a/q2_feature_table/_subsample.py
+++ b/q2_feature_table/_subsample.py
@@ -15,6 +15,11 @@ def subsample(table: biom.Table, subsampling_depth: int,
         # we are transposing the table due to biocore/biom-format#759
         table = table.transpose()
 
+    if len(table.ids()) < subsampling_depth:
+        raise ValueError('The subsampling depth exceeds the number of '
+                         'elements on the desired axis. The maximum depth '
+                         'is: %d.' % len(table.ids()))
+
     # the axis is always 'sample' due to the above transpose
     table = table.subsample(subsampling_depth, axis='sample', by_id=True)
 
@@ -25,5 +30,10 @@ def subsample(table: biom.Table, subsampling_depth: int,
     if axis == 'feature':
         # reverse the transpose necessary due to biocore/biom-format#759
         table = table.transpose()
+
+    if table.is_empty():
+        raise ValueError('The subsampled table contains no samples or '
+                         'features. Verify your table is valid and that you '
+                         'provided a shallow enough subsampling depth.')
 
     return table
--- a/q2_feature_table/_subsample.py
+++ b/q2_feature_table/_subsample.py
@@ -32,8 +32,9 @@ def subsample(table: biom.Table, subsampling_depth: int,
         table = table.transpose()
 
     if table.is_empty():
-        raise ValueError('The subsampled table contains no samples or '
-                         'features. Verify your table is valid and that you '
-                         'provided a shallow enough subsampling depth.')
+        raise ValueError('The subsampled table contains no samples or features'
+                         ' (samples/features that sum to zero after filtering'
+                         ' are automatically removed). It may be a good idea'
+                         ' to double check that your table is valid/nonempty.')
 
     return table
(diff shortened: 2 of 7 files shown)


wasade:
...and note, this is impacted by #biocore/biom-format#759 but the work around is not terrible. On a related note.

jairideout:
I'll work on getting biom-format updated to 2.1.6 (#144), will follow up here when it's ready.

jairideout:
biom-format 2.1.6 is now part of the [latest QIIME 2 environment files](https://github.com/qiime2/environment-files/tree/master/latest/staging)! This means that local dev environments created from the latest staging file with include the updated biom-format package, and Travis-CI will also use that package when testing.

@wasade can you update this PR as necessary for biom-format 2.1.6? As part of that, please change the version requirement in `ci/recipe/meta.yaml` to `biom-format >=2.1.6,<2.2.0`. Let me know when this PR is ready for review. Thanks!

wasade:
Thanks! I don't think there are changes necessary for 2.1.6 short of `meta.yaml` which I didn't realize was contained in the repo. Awesome

jairideout:
Do the changes in this PR *require* biom-format 2.1.6? If not, the version bump in `meta.yaml` should be reverted.

jairideout:
Can you revert this commit? https://github.com/qiime2/q2-feature-table/pull/146/commits/ce60328c63b7c482c3c77f84fea2c48f3895a71a

q2_feature_table/_sample.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import biom
+
+
+def sample(table: biom.Table, sampling_depth: int, axis: str) -> biom.Table:
```
jairideout: The current API naming could be confusing because the feature table has samples, so this verbiage suggests that users *sample the samples* (if `axis='sample'`). The parameter `sampling_depth` is also used in a different context within qiime2 (i.e. for rarefaction depth).

@gregcaporaso and I chatted about this API, what do you think?

```python
subsample(table, n, axis):
```

wasade: :+1: agree it was a bit confusing even if accurate :) Will revise in a moment

q2_feature_table/tests/test_subsample.py
```diff
@@ -0,0 +1,56 @@
...
+        sample_ids = frozenset(a.ids(axis='observation'))
+        self.assertIn(sample_ids, set([frozenset(['O1', 'O2']),
+                                       frozenset(['O1', 'O3']),
+                                       frozenset(['O2', 'O3'])]))
+        self.assertEqual(set(a.ids(axis='sample')), set(['S1', 'S2']))
+
+        for i in a.ids(axis='observation'):
+            npt.assert_equal(t.data(i, axis='observation'),
+                             a.data(i, axis='observation'))
+
```
ebolyen: I think there's a few more edge cases to test:

```
   f1
s1 0
s2 1
```
where `subsample_depth=1`, `axis='sample'` and `s1` is picked. Then the same situation for the other axis.
This should probably error since there wouldn't be any features/samples left in the table.

Then if the axis you are sampling has no ids, then the `sampling_depth <= ids` should prevent that from succeeding, but it's still possible for the opposing axis to be null. If the above edge cases error, then I would probably expect this situation to error as well, because the end result is an empty table along an axis.

jairideout: :+1: for erroring whenever an empty table would be produced (we're trying to disallow empty feature tables in qiime2, there's an open issue somewhere to make that a reality someday)

q2_feature_table/_subsample.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def subsample(table: biom.Table, subsampling_depth: int,
+              axis: str) -> biom.Table:
+    if axis == 'feature':
+        # we are transposing the table due to biocore/biom-format#759
+        table = table.transpose()
+
+    # the axis is always 'sample' due to the above transpose
+    table = table.subsample(subsampling_depth, axis='sample', by_id=True)
```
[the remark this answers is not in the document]
wasade: All IDs are [retained](https://github.com/biocore/biom-format/blob/master/biom/table.py#L2759)

q2_feature_table/plugin_setup.py
```diff
@@ -54,6 +54,30 @@
 )
 
 plugin.methods.register_function(
+    function=q2_feature_table.subsample,
+    inputs={'table': FeatureTable[Frequency]},
+    parameters={'subsampling_depth': Int,
```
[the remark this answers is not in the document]
wasade: Oh, nifty... adding

q2_feature_table/plugin_setup.py
```diff
@@ -54,6 +54,30 @@
...
+        'subsampling_depth': ('The total number of samples or features to be '
+                              'randomly sampled. Samples or features that are '
+                              'reduced to a zero sum will not be included in '
+                              'the resulting table.'),
+        'axis': ('The axis to sample over. If "sample" then samples will be '
+                 'randomly selected to be retained. If "feature" then '
+                 'a random set of features will be selected to be retained.')
+    },
+    output_descriptions={
+        'sampled_table': 'The resulting sampled feature table.'
```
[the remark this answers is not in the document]
wasade: Done

q2_feature_table/plugin_setup.py
```diff
@@ -54,6 +54,30 @@
...
+    },
+    name='Sample table',
```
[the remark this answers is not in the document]
wasade: Done

q2_feature_table/tests/test_subsample.py
```diff
@@ -0,0 +1,56 @@
...
+        sample_ids = frozenset(a.ids(axis='sample'))
+        self.assertIn(sample_ids, set([frozenset(['S1', 'S2']),
+                                       frozenset(['S1', 'S3']),
+                                       frozenset(['S2', 'S3'])]))
+        self.assertEqual(set(a.ids(axis='observation')), set(['O1', 'O2']))
+
+        for i in a.ids(axis='sample'):
+            npt.assert_equal(t.data(i, axis='sample'),
+                             a.data(i, axis='sample'))
+
```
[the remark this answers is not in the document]
wasade: Done

q2_feature_table/tests/test_subsample.py
wasade: Will QIIME allow for features or observations to be empty as in the above example? Ideally they're removed the the table since they're null.

`Table.subsample` returns empty tables without raising an exception; this isn't a problem for the underlying API. We can, however, test and raise an error within the plugin method. What error do you suggest?

ebolyen: @wasade, what about something along the lines of 
`Subsampling has removed all features from the table (0-sum features are automatically removed)` and then similar for `samples` too.

ebolyen: Also, since it looks like you can have a depth greater than the number of elements in an axis, we should probably guard the input for null axes (and test that too). Alternatively, `subsample` could choose to error if the `subsample_depth` exceeds the number of elements in an axis.

I'm fine with either approach.

wasade: Cool cool, I think those last changes resolve it?

q2_feature_table/_subsample.py
```diff
@@ -0,0 +1,39 @@
...
+    table.filter(lambda v, i, m: v.sum() > 0, axis=invaxis)
+
+    if axis == 'feature':
+        # reverse the transpose necessary due to biocore/biom-format#759
+        table = table.transpose()
+
+    if table.is_empty():
+        raise ValueError('The subsampled table contains no samples or '
+                         'features. Verify your table is valid and that you '
+                         'provided a shallow enough subsampling depth.')
```
ebolyen: `contains either no samples or features` (since there's probably still one non-null axis, otherwise the subsampling_depth check would have tripped up first).

And then the second sentence seems a little backwards, wouldn't it be the shallow subsampling depths that are more likely to grab result in a null axis? Maybe explaining the `table.filter` call that happens would be a better hint as to what happens:

```python
"The subsampled table contains either no sample or features (samples/features that sum to zero are automatically removed)."

ebolyen:
@wasade, ping on the lint error. I can also push the changes and merge if you don't have time.

PR #156, opened 2017-11-30
Prevent Negative Slider Value

### LINKED ISSUE #125: disallow negative sampling depth in ticker
If using the ticker to select a sampling depth or typing in a number manually (i.e. *not* using the slider), negative sampling depths can be entered:

image

patthehat033:
I modified the sample-frequency-detail.html file and made the slider-value have a minimum of 0.  I also modified the controls.js and made the sliderValue = 0 if the sliderValue is changed to a negative number. 

Fixes #125

### DIFF
--- a/q2_feature_table/_summarize/summarize_assets/sample-frequency-detail.html
+++ b/q2_feature_table/_summarize/summarize_assets/sample-frequency-detail.html
@@ -12,6 +12,7 @@
         id="slider-value"
         class="form-control"
         type="number"
+        min="0"
       >
       <div>(Zero implies no even sampling.)</div>
       <br>
--- a/q2_feature_table/_summarize/summarize_assets/src/controls.js
+++ b/q2_feature_table/_summarize/summarize_assets/src/controls.js
@@ -61,15 +61,14 @@ const initializeControls = () => {
     if (+sliderValue.node().value > +slider.node().max) {
       sliderValue.node().value = slider.node().max;
       slider.node().value = slider.node().max;
-    } else {
-      slider.node().value = sliderValue.node().value;
     }
-
-    if (sliderValue.node().value < 0) {
+    else if (sliderValue.node().value < 0) {
       sliderValue.node().value = 0;
       slider.node().value = slider.node().min;
     }
-
+    else {
+      slider.node().value = sliderValue.node().value;
+    }
     updateStats(trows, slider);
     toggleStyles(trows, slider);
     hiddenDepth.attr('value', slider.node().value).dispatch('change');
--- a/q2_feature_table/_summarize/summarize_assets/src/controls.js
+++ b/q2_feature_table/_summarize/summarize_assets/src/controls.js
@@ -70,7 +70,7 @@ const initializeControls = () => {
   });
 
   sliderValue.on('change', () => {
-    if (!sliderValue.node().value) {
+    if (!sliderValue.node().value || sliderValue.node().value < 0) {
       sliderValue.node().value = 0;
       slider.node().value = slider.node().min;
     }
--- a/q2_feature_table/_summarize/summarize_assets/src/controls.js
+++ b/q2_feature_table/_summarize/summarize_assets/src/controls.js
@@ -64,13 +64,19 @@ const initializeControls = () => {
     } else {
       slider.node().value = sliderValue.node().value;
     }
+
+    if (sliderValue.node().value < 0) {
+      sliderValue.node().value = 0;
+      slider.node().value = slider.node().min;
+    }
+
     updateStats(trows, slider);
     toggleStyles(trows, slider);
     hiddenDepth.attr('value', slider.node().value).dispatch('change');
   });
 
   sliderValue.on('change', () => {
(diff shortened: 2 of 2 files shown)


thermokarst:
Thanks @patthehat033 - we are currently in a feature freeze while we get QIIME 2 2017.11 out the door, but we expect to be accepting new PRs within the next week or so (we will remove the `DO-NOT-MERGE` label when we are unfrozen). Thanks!

q2_feature_table/_summarize/summarize_assets/sample-frequency-detail.html
```diff
@@ -12,6 +12,7 @@
         id="slider-value"
         class="form-control"
         type="number"
+        min="0"
```
thermokarst: 👍

q2_feature_table/_summarize/summarize_assets/src/controls.js
```diff
@@ -70,7 +70,7 @@ const initializeControls = () => {
   });
 
   sliderValue.on('change', () => {
-    if (!sliderValue.node().value) {
+    if (!sliderValue.node().value || sliderValue.node().value < 0) {
```
thermokarst: I think you are off to a great start, but it doesn't quite solve the [original issue](https://github.com/qiime2/q2-feature-table/issues/125):

> If using the ticker to select a sampling depth or typing in a number manually (i.e. not using the slider), negative sampling depths can be entered

Specifically, when entering a number manually, the input still allows you to enter a negative value. I think the changes you have here need to be moved up into the `input` handler [beginning on line 60](https://github.com/qiime2/q2-feature-table/pull/156/files#diff-eb122d510b1bc91b43ea49a0511ebbd6R60). You will probably need to add another branch in the `if/else` flow here to check if the value is negative.

patthehat033: I agree the input handler seems like a better place to put it!  I modified the code.

thermokarst: Thanks @patthehat033!

> You will probably need to add another branch in the if/else flow here to check if the value is negative.

Can you add this code as a new branch in the if/else - that will probably make more sense, versus a standalone check. How about:

```javascript
if (entered value is larger than the max slider value) {
    ...
else if (entered value is less than zero) {
    ...
else {
    ...
}
```

thermokarst: Sorry, sent too soon! The reason we would want this in the existing branching, versus a new check, is because as you have it at the moment, this will allow a negative value to exist, albeit briefly. If you add the check as a new branch in the validation, then you can ensure we never enter an inconsistent state. Make sense?

patthehat033: @thermokarst I updated the code!  Although, I am a little confused.  I know that having it as a standalone branch may be more inefficient since the if statement for checking if the value is negative would always be executed even if the number is greater than the max slider value.  In terms of allowing a negative number to exist briefly though, wouldn't the logic still be the same for the computer? It would check the first if statement (the greater than max slider value) and then check the if statement for the negative number (or else if in this case)?  Just curious how this works.

thermokarst: Good question @patthehat033! So if we left it as you had it in your [previous revision](https://github.com/qiime2/q2-feature-table/pull/156/commits/3f99c7bbdd1e9e86dbcc3d1e662b689d31c3b0a2), we can walk through a brief example:

1) The user enters a negative value. Entering text in the input field triggers the [`on-input` handler](https://github.com/qiime2/q2-feature-table/pull/156/commits/3f99c7bbdd1e9e86dbcc3d1e662b689d31c3b0a2#diff-eb122d510b1bc91b43ea49a0511ebbd6R60).

2) Since the value is negative, the [first branch](https://github.com/qiime2/q2-feature-table/pull/156/commits/3f99c7bbdd1e9e86dbcc3d1e662b689d31c3b0a2#diff-eb122d510b1bc91b43ea49a0511ebbd6R61) is skipped, which takes us to the [second branch](https://github.com/qiime2/q2-feature-table/pull/156/commits/3f99c7bbdd1e9e86dbcc3d1e662b689d31c3b0a2#diff-eb122d510b1bc91b43ea49a0511ebbd6R64).

3) This branch naively sets the **slider** to match whatever the negative value is. This triggers the slider's `on-change` event to fire, even though we are still handling the first event. This isn't a problem now, because we don't have any custom actions bound to the slider's `on-change`, but, can you imagine what would happen if we did add some event code here? We could run into a situation where it is difficult to track down what is happening, and why.

At the end of the day, I like having this as an explicit branch in the `if/else`, not because it is more performant, but because I think it is very clear for any future developers to understand what is happening here, and why. Make sense? Sometimes we have to make these kinds of decisions for the *people* that will be reading the code, versus the *computers* that will be executing them.

patthehat033: @thermokarst Thank you for the well written explanation!  That completely makes sense!  I was overlooking the fact that the slider would be affected even if the number was negative.

PR #159, opened 2017-12-19
BUG: Disable feature freq. rugplot

### LINKED ISSUE #122: Performance issue with summarize and large tables
I am running the summarize visualizer on a table with ~300,000 features and 120 samples. This has been running for about 2 hours now. I would have not expected this command to take so long.

Specific code:

```python
import qiime2 as q2
from qiime2.plugins import feature_table

feature_table.visualizers.summarize(q2.Artifact.load('feature-table.qza'),
                                    q2.Metadata.load('mapping-file.tsv'))
```

After talking with @ebolyen and looking at the `TMPDIR`, it seems like the visualizer is getting hung on the feature-level summary.

thermokarst:
Fixes #122 

On a test table with 65k features, this brought the runtime down to ~25 seconds, from ~35 minutes previously. The bottleneck appears to be matplotlib's rendering of the rugplot, so by disabling, we get a bit of performance gain more or less for free. The data is still available as a delimited download in the viz, so if users are interested in more complex figures, they have the data available to make it happen!

### DIFF
--- a/q2_feature_table/_summarize/_visualizer.py
+++ b/q2_feature_table/_summarize/_visualizer.py
@@ -81,7 +81,7 @@ def summarize(output_dir: str, table: biom.Table,
         table, axis='observation')
     if number_of_features > 1:
         feature_frequencies_ax = sns.distplot(feature_frequencies, kde=False,
-                                              rug=True)
+                                              rug=False)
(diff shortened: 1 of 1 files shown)


gregcaporaso:
Thanks @thermokarst! This will be a much-appreciated improvement!

PR #162, opened 2018-01-03
PERF: prefer Table.concat over Table.merge

wasade:
Modifies `q2_feature_table.merge` to prefer the use of `Table.concat` which is 1-2 orders magnitude faster than `Table.merge` with increasing performance gains as the number of tables to merge gets large.

### DIFF
--- a/q2_feature_table/_merge.py
+++ b/q2_feature_table/_merge.py
@@ -11,16 +11,16 @@ import pandas as pd
 import collections
 
 
+def overlap_methods():
+    return ('error_on_overlapping_sample', 'error_on_overlapping_feature',
+            'sum')
+
+
 def _get_overlapping(tables, axis):
     ids = collections.Counter()
     for table in tables:
         ids.update(table.ids(axis=axis))
     return {e for e, c in ids.items() if c > 1}
-
-
-def overlap_methods():
-    return ('error_on_overlapping_sample', 'error_on_overlapping_feature',
-            'sum')
 
 
 def merge(tables: biom.Table,
--- a/q2_feature_table/_merge.py
+++ b/q2_feature_table/_merge.py
@@ -11,6 +11,13 @@ import pandas as pd
 import collections
 
 
+def _get_overlapping(tables, axis):
+    ids = collections.Counter()
+    for table in tables:
+        ids.update(table.ids(axis=axis))
+    return {e for e, c in ids.items() if c > 1}
+
+
 def overlap_methods():
     return ('error_on_overlapping_sample', 'error_on_overlapping_feature',
             'sum')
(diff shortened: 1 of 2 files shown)


thermokarst:
Thanks @wasade! Looks like travis is picking up the old travis config (we [recently](https://forum.qiime2.org/t/new-conda-installation-scheme/2528/8) had to tweak how we deploy conda env files) - I imagine a quick rebase would get this moving in the right direction. Thanks!

q2_feature_table/_merge.py
```diff
@@ -16,32 +16,28 @@ def overlap_methods():
...
-                             % ', '.join(overlapping_ids))
-    elif overlap_method == 'error_on_overlapping_feature':
-        overlapping_ids = _get_overlapping(tables, 'observation')
-        if len(overlapping_ids) > 0:
-            raise ValueError('Same features are present in provided tables: %s'
-                             % ', '.join(overlapping_ids))
+    if len(tables) == 1:
+        return tables[0]
+
+    if overlap_method.startswith('error_on_overlapping'):
```
thermokarst: Do we need this top-level `if`? Seems like if we dropped it, then we don't need to raise the `Invalid overlap method` [exception](https://github.com/qiime2/q2-feature-table/pull/162/files#diff-b32385f4b690efbf967cf6f0d87609b9R30) in [two](https://github.com/qiime2/q2-feature-table/pull/162/files#diff-b32385f4b690efbf967cf6f0d87609b9R37) different places.

```python
if overlap_method == 'error_on_overlapping_sample':
    ...
elif overlap_method == 'error_on_overlapping_feature':
    ...
elif overlap_method == 'sum':
    ...
else:
    ...
```

q2_feature_table/_merge.py
```diff
@@ -16,32 +16,28 @@ def overlap_methods():
...
+            return tables[0].concat(tables[1:], 'sample')
+        elif overlap_method == 'error_on_overlapping_feature':
+            return tables[0].concat(tables[1:], 'observation')
+        else:
+            raise ValueError('Invalid overlap method: %s. Please provide one of '
+                             'the following methods: %s.' %
+                             (overlap_method, ', '.join(overlap_methods())))
     elif overlap_method == 'sum':
         # This is the default behavior for biom.Table.merge
         pass
```
thermokarst: Since this PR proposes early returns in the other branches of this `if/else`, how about you move [the rest of the `sum` logic](https://github.com/qiime2/q2-feature-table/pull/162/files#diff-b32385f4b690efbf967cf6f0d87609b9R46) up into here, that way the early returns are consistent.

q2_feature_table/tests/test_merge.py
```diff
@@ -62,7 +63,8 @@ def test_invalid_overlapping_feature_ids(self):
         t2 = Table(np.array([[0, 2, 6], [2, 2, 4]]),
                    ['O1', 'O3'],
                    ['S4', 'S5', 'S6'])
-        with self.assertRaisesRegex(ValueError, 'features are present'):
+        with self.assertRaisesRegex(biom.exception.DisjointIDError,
```
thermokarst: Would it make sense to trap the exception raised by `biom` in the method, that way we can maintain the same error message? I think it might be nice to maintain the same error message that users have already been exposed to - what do you think?

q2_feature_table/_merge.py
```diff
@@ -16,32 +16,28 @@ def overlap_methods():
...
-        if len(overlapping_ids) > 0:
-            raise ValueError('Same samples are present in provided tables: %s'
-                             % ', '.join(overlapping_ids))
-    elif overlap_method == 'error_on_overlapping_feature':
-        overlapping_ids = _get_overlapping(tables, 'observation')
-        if len(overlapping_ids) > 0:
-            raise ValueError('Same features are present in provided tables: %s'
-                             % ', '.join(overlapping_ids))
+    if len(tables) == 1:
+        return tables[0]
```
thermokarst: Cool, I like this short-circuit.

q2_feature_table/_merge.py
wasade: Good call, I like this approach

q2_feature_table/_merge.py
wasade: Thanks!

q2_feature_table/_merge.py
wasade: Agree, good call

q2_feature_table/tests/test_merge.py
wasade: Okay, trapping. It does obfuscate the traceback, unless we pull out the st but I'm not sure if the increased complexity of the error handling is sufficiently beneficial.

wasade:
Thanks for the review @thermokarst, I like your approach here. Let me know if more changes are needed

wasade:
I'm a bit confused, I was not observing these CI errors when testing locally before pushing, and now when I load what was a fresh q2 environment as of a an hour ago, I'm getting the following traceback:

```python
Traceback (most recent call last):
  File "q2_feature_table/tests/test_merge.py", line 17, in <module>
    from q2_feature_table import merge, merge_seqs, merge_taxa
  File "/Users/mcdonadt/ResearchWork/software/q2-feature-table/q2_feature_table/__init__.py", line 16, in <module>
    from ._group import group
  File "/Users/mcdonadt/ResearchWork/software/q2-feature-table/q2_feature_table/_group.py", line 41, in <module>
    metadata: qiime2.CategoricalMetadataColumn, mode: str) -> biom.Table:
AttributeError: module 'qiime2' has no attribute 'CategoricalMetadataColumn'
```

thermokarst:
Hmm, sounds like something went haywire! Is `qiime2` installed in development mode in this env? If so, have you fetched the latest changes? We had a sweeping change across all repos last week related to a major metadata refactor.

wasade:
What's development mode...? I did a fresh install this morning of 2017.12 (e.g., these [instructions](https://docs.qiime2.org/2017.12/install/native/#install-qiime-2-within-a-conda-environment)), and then a `pip install -e .` of this repo

thermokarst:
Awesome, thanks @wasade! I pushed up a [tiny commit](https://github.com/qiime2/q2-feature-table/pull/162/commits/ec2d7b4fe4f25623c44a72d200223ad849f0eb94) to clean up the diff a bit - please let me know if that is okay, and if not, feel free to nuke that commit and we can come up with another strategy. Thanks!

wasade:
Totally okay, thank you!!

PR #190, opened 2018-10-01
add with_replacement to rarefy

stevendbrown:
Added `with_replacement` option to `rarefy`, documentation to this effect, and a test that it should return the correct shape when `with_replacement` is set to `True`.

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -22,7 +22,7 @@ requirements:
     - python 3.5*
     - setuptools
     - scikit-bio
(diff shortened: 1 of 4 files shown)


stevendbrown:
I've fixed the flake8 errors, but I'd rather not clutter up the commit history with those fixes until I can fix the more substantive issues, which seem to be that `biom` doesn't have `with_replacement` as a parameter to `subsample`.

stevendbrown:
Closing to split into two PRs.

PR #199, opened 2019-01-02
DOC: BIOM citation

wasade:
The BIOM-Format citation wasn't included in the citations file. This pull request simply adds the .bib entry from GigaScience. This PR does not currently use the citation on any action as it wasn't clear to me where it should go.

### DIFF
--- a/q2_feature_table/citations.bib
+++ b/q2_feature_table/citations.bib
@@ -1,3 +1,30 @@
+@Article{McDonald2012,
+author="McDonald, Daniel
+and Clemente, Jose C.
+and Kuczynski, Justin
+and Rideout, Jai Ram
+and Stombaugh, Jesse
+and Wendel, Doug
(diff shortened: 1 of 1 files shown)


ebolyen:
This citation is attached to the view/formats in q2-types:
https://github.com/qiime2/q2-types/blob/c8c0f1ff426d31a4630f5519566c3551921dbb65/q2_types/feature_table/_format.py#L95-L97

which means it will be tracked anytime one of those views is used in an action/transformer (which for this plugin is basically all of them).

thermokarst:
I agree w/ @ebolyen  --- all of the Actions in this plugin that utilize `biom` views will already have this citation tacked on --- we have hit citation bedrock. @wasade - how do you feel about closing this PR?

wasade:
That is awesome! All for closing, thanks!

PR #206, opened 2019-04-04
allow heatmap to not cluster on samples or features

### LINKED ISSUE #173: Allow for no clustering on either axis of heatmap viz
The `heatmap` viz supports axis clustering when plotting - this issue proposes adding a new option to the existing available options.

**Current Behavior**
Current options are to cluster samples, features, or both samples and features (options `samples`, `features`, `both`, respectively).

**Proposed Behavior**
It would be nice to also support generating this viz with no clustering applied to the axes (`cluster='none'`). This would render the IDs in the same order that they are in in the `biom.Table` used to generate the viz.

**Questions**
1. The current default for the `cluster` parameter is `both`, and it seems like it makes sense to leave that value unchanged with the addition of the `none` option.

**References**
1. [Original forum post](https://forum.qiime2.org/t/feature-table-heatmap-with-no-clustering-on-neither-axes/3826?u=nicholas_bokulich)
2. [Another request on the forum](https://forum.qiime2.org/t/heat-map-issues/3980)

gwarmstrong:
Fixes #173 

Updates: _heatmap options for clustering, options and docs for plugin, unit test for feature

---

Sample output with no clustering (`none` option).

Example-none-Cluster

### DIFF
--- a/q2_feature_table/plugin_setup.py
+++ b/q2_feature_table/plugin_setup.py
@@ -469,8 +469,7 @@ plugin.visualizers.register_function(
         'method': 'Clustering methods exposed by seaborn (see http://seaborn.'
                   'pydata.org/generated/seaborn.clustermap.html#seaborn.clust'
                   'ermap for more detail).',
-        'cluster': 'Specify which axes to cluster. Option to cluster on '
-                   'samples, features, both or none',
+        'cluster': 'Specify which axes to cluster.',
         'color_scheme': 'The matplotlib colorscheme to generate the heatmap '
                         'with.',
     },
--- a/q2_feature_table/plugin_setup.py
+++ b/q2_feature_table/plugin_setup.py
@@ -469,7 +469,8 @@ plugin.visualizers.register_function(
         'method': 'Clustering methods exposed by seaborn (see http://seaborn.'
                   'pydata.org/generated/seaborn.clustermap.html#seaborn.clust'
                   'ermap for more detail).',
-        'cluster': 'Specify which axes to cluster.',
+        'cluster': 'Specify which axes to cluster. Option to cluster on '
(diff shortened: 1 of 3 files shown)


q2_feature_table/plugin_setup.py
```diff
@@ -469,7 +469,8 @@
         'method': 'Clustering methods exposed by seaborn (see http://seaborn.'
                   'pydata.org/generated/seaborn.clustermap.html#seaborn.clust'
                   'ermap for more detail).',
-        'cluster': 'Specify which axes to cluster.',
+        'cluster': 'Specify which axes to cluster. Option to cluster on '
```
thermokarst: Can you drop the `Option to cluster on...` text? Interfaces will render the options out, instead:

```bash
# q2cli
  --p-cluster [features|both|samples]
                                  Specify which axes to cluster.  [default:
                                  both]
```

```python
# Artifact API
cluster : Str % Choices('features', 'both', 'samples'), optional
    Specify which axes to cluster.
```

gwarmstrong: Sure thing.
To clarify, the q2cli and Artifact API options shown will be updated automatically based on the options available in `q2_feature_table/_heatmap/_visualizer.py`?

thermokarst: Yep!

thermokarst: QIIME 2 knows about those `Choices` via the method's registration: https://github.com/qiime2/q2-feature-table/blob/master/q2_feature_table/plugin_setup.py#L445

gwarmstrong: Awesome, thanks for the link to the relevant code!

PR #212, opened 2019-05-02
ENH: filter-samples supports multiple types

ebolyen:


q2_feature_table/plugin_setup.py
```diff
@@ -236,17 +236,19 @@
                 "the data from the first will be propagated to the result."
 )
 
+T = TypeMatch([Frequency, RelativeFrequency, PresenceAbsence, Composition])
+# TODO: constrain min/max frequency when optional is handeled by typemap
```
thermokarst:
```suggestion
# TODO: constrain min/max frequency when optional is handled by typemap
```

q2_feature_table/_filter.py
```diff
@@ -56,8 +56,8 @@ def _filter_table(table, min_frequency, max_frequency, min_nonzero,
     # filter on the opposite axis to remove any entities that now have a
     # frequency of zero
     filter_fn2 = _get_biom_filter_function(
-        ids_to_keep=table.ids(axis=_other_axis_map[axis]), min_frequency=1,
-        max_frequency=None, min_nonzero=0, max_nonzero=None)
+        ids_to_keep=table.ids(axis=_other_axis_map[axis]), min_frequency=0,
+        max_frequency=None, min_nonzero=1, max_nonzero=None)
```
ebolyen: This was breaking filtering an relative-frequency (as all values were less than one). Using min_nonzero of 1 gets to the point better and doesn't assume a dtype.

PR #218, opened 2019-11-05
ENH: Filter bloom sequences

gwarmstrong:
**New Feature**
* Filter microbial blooms from feature tables

**Importance**
[Amir et al. (2019)](https://msystems.asm.org/content/2/2/e00199-16) have recently shown that  several bacterial taxa grow at room temperature, which can confound analyses on some subsets of AGP data. A recommended practice for handling this issue is filtering [these](https://github.com/knightlab-analyses/bloom-analyses/blob/master/data/newbloom.all.fna) from your feature table as show in [this](https://github.com/knightlab-analyses/bloom-analyses/blob/master/ipynb/bloom_example.ipynb) tutorial.

Discussing with @wasade it seemed that adding this into Q2 might be nice, so here is my first shot at an addition.

**Usage**
You could do something like

```bash
qiime feature-table filter-bloom-features \
    --i-table table.qza \
    --i-data sequences.qza \
    --o-filtered-table bloom-filtered-table.qza
```

And if you needed something like an updated version of the bloom sequences published stored as `custom-bloom-seqs.qza`

You could do something like 
```bash
qiime feature-table filter-bloom-features \
    --i-table table.qza \
    --i-data sequences.qza \
    --i-bloom-sequences custom-bloom-seqs.qza \
    --o-filtered-table bloom-filtered-table.qza
```

---

**Include any questions for reviewers, screenshots, sample outputs, etc.**

* Does this like a reasonable addition for `q2-feature-table`?
* I am not sure if I updated the `package_data` in the `setup.py` correctly to account for the sequences I included in the package.
* I tried to handle the assets based off of other Q2 plugins. Does the way its set up for the default sequences seem fine?
* Would you rather the sequences were uploaded as a Q2 artifact? I figured uploaded the FASTA may be a little more transparent.


ebolyen:
Hey @gwarmstrong!

Thanks for this PR, I think there's two things that are stopping me from merging, but altogether the code itself looks excellent.

1. We really want to avoid vendoring data like the bloom-filter. This should be an artifact that users can provide (so that they can pick their filter or adapt it). We are more than happy to host an artifact such as this, we just believe that it should be an explicit step that the users chooses to take.

2. If the sequences became an artifact, then we already have an action that does this (and a bit more): [`quality-control exclude-seqs`](https://docs.qiime2.org/2019.10/plugins/available/quality-control/exclude-seqs/), granted it doesn't change the table, but that could be done with a second action. So I could see a really nice pipeline that uses a few of these filters to "clean up" your artifacts, and this pipeline would make sense in q2-quality-control I think.

Perhaps a data-resource, new tutorial, and convenience pipeline would be the best way to go about this?

gwarmstrong:
Hi @ebolyen 

Thanks for the feedback! Makes sense on both reservations. I like the data-resource, new tutorial, convenience pipeline approach. Feel free to close out this PR.

Re:
1. What is the process for submitting/posting data resources?

2. I'll write up something in up inside of `q2-quality-control` that takes a `table: FeatureTable[Frequency]`, `data: FeatureData[Sequence]`, `exclude-seqs FeatureData[Sequence]`, and outputs filtered version of `table` and `data` without features that are in `exclude-seqs`. Sound good?

Tutorial: I can submit a draft of the tutorial in the description of the `q2-quality-control` PR?

Thanks!

ebolyen:
Thanks @gwarmstrong!

Re: 1, there isn't one, if you send us an artifact, we'll upload to S3 and update our permalinks appropriately. This is something the library will one day manage, but for the moment its a manual process.

Re: 2, yep, that sounds perfect!

Re: tutorial, that also sounds good, once the content exists, we can figure out a nice place for it.

Issue #223, opened 2020-02-28
Add method to re-name features or samples in feature table


nbokulich:
**Addition Description**
Given a sample metadata file/column and a feature table, re-name the feature table to use the labels in the specified column instead.

**Current Behavior**
This can be achieved with `feature_table.group()` in a roundabout way — and it is slow since the method is grouping and summing instead of just re-indexing. Having a dedicated method to this often-requested feature (based on forum requests) would be ideal.

**Proposed Behavior**
`biom.Table.update_ids()` the chosen feature table axis (sample/feature) using the provided metadata. Need to add some basic validation/processing:
1. ensure that all values in the metadata column are unique
2. re-order the metadata index to match the order in the table

**Questions**
1. What if metadata is a superset? Probably just ignore the extra IDs.
2. What if metadata is a subset? error or only relabel that subset of IDs? (I think both: add a parameter to adjust missing ID handling, which would be error by default)

jwdebelius:
I know we talked about this a month ago, and I want to test the waters with plugins before I try to build my own. So, one question is whether we should take a metadata column as input or just a tsv kind of map.

thermokarst:
I think it should be a metadata column (which QIIME 2 interfaces like q2cli can extract from a metadata tsv file).

PR #225, opened 2020-03-25
Rename table

jwdebelius:
This addresses #223. It creates a renaming function based on a metadata column.

Outstanding:
There is theoretically a flag to add a warning (currently called "verbose") which would list off the samples that are being excluded/skipped. This can either be dropped, or needs to be hooked into the verbose flag in the plugin.

### DIFF
--- a/q2_feature_table/_rename.py
+++ b/q2_feature_table/_rename.py
@@ -54,14 +54,17 @@ def _generate_new_names(old_ids, rename, strict, verbose=False):
 
 def rename_samples(table: biom.Table, 
                    metadata: qiime2.CategoricalMetadataColumn, 
+                   axis: str = 'sample',
                    strict: bool = False)\
                     -> biom.Table:
     
     rename = metadata.to_series()
-    old_ids = table.ids(axis='sample')
+    if axis == 'feature':
+        axis = 'observation'
+    old_ids = table.ids(axis=axis)
 
     new_ids = _generate_new_names(old_ids, rename, strict, False)
 
-    updated = table.update_ids(new_ids, axis='sample', inplace=False)
+    updated = table.update_ids(new_ids, axis=axis, inplace=False)
 
     return updated
--- a/q2_feature_table/_rename.py
+++ b/q2_feature_table/_rename.py
@@ -10,8 +10,6 @@ import warnings
 
 import biom
 import qiime2
-import numpy as np
-import pandas as pd
 
 
 def _generate_new_names(old_ids, rename, strict, verbose=False):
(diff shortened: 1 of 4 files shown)


q2_feature_table/_rename.py
```diff
@@ -0,0 +1,71 @@
...
+            'There are ids supplied for renaming that are not in the table.\n'
+            'The following ids will not be mapped:\n'
+            '   %s' % missing_rename)
+        )
+
+    new_ids = {id_: rename.to_dict().get(id_, id_) for id_ in old_ids}
+    return new_ids
+
+
+def rename_samples(table: biom.Table,
```
thermokarst, line 56: Since this method can rename samples && features, what if we call it something like `relabel_ids` (or something along those line)? Maybe `rename_ids` - that way the use of the word "name" in the method helptext still makes sense?

q2_feature_table/_rename.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def _generate_new_names(old_ids, rename, strict, verbose=False):
+    """
+    Checks the list of samples ot be renamed
+    """
+    # all old ids need to be unique
+    if (rename.value_counts() > 1).any():
+        raise ValueError('All new ids must be unique.\n'
+                         'Try qiime feature-table group if you want '
```
thermokarst, line 22: I like where you're going with this, but the formatting of the command might be confusing, since it is hardcoded as a q2cli invocation. I think in a case like this you can get away with just referring to the "method `group` on this plugin", or something along those lines.

q2_feature_table/_rename.py
```diff
@@ -0,0 +1,71 @@
...
+                         'to combine multipel samples in the same table.')
```
thermokarst, line 23:
```suggestion
                         'to combine multiple samples in the same table.')
```

Also, related to my other comment about samples vs features - can we come up with another word to use here, instead of samples?

q2_feature_table/plugin_setup.py
```diff
@@ -236,6 +236,41 @@
...
+    inputs={
+        'table': FeatureTable[Frequency | RelativeFrequency],
+    },
+    parameters={
+        'metadata': MetadataColumn[Categorical],
+        'strict': Bool,
+        'axis': Str % Choices({'sample', 'feature'})
+        },
+    outputs=[
+        ('renamed_table', FeatureTable[Frequency | RelativeFrequency])
```
thermokarst, line 250: I like where you're going with this, but we'll need to use a `TypeMatch` in order to make this work:

```python
T = TypeMatch([Frequency, RelativeFrequency])

    ...
    inputs={'table': FeatureTable[T]},
    ...
    outputs={'renamed_table': FeatureTable[T]},
    ...
)
```

This tells QIIME 2 to return a `FeatureTable[Frequency]` when it receives a `FeatureTable[Frequency]` as input, and similarly with `RelativeFrequency`.

You could also expand that `TypeMatch` to include all kinds of other feature tables subtypes

```python
TypeMatch([Frequency, RelativeFrequency, PresenceAbsence, Composition])
```

jwdebelius: I didn't know this was possible. super cool!

q2_feature_table/_rename.py
jwdebelius: I updated as suggested, although without the formatting. I don't think my CLI interpreter can handle the back ticks. Can you do a **bold** in command line arguments easily?

q2_feature_table/_rename.py
jwdebelius: I think I replaced this with ids

jwdebelius:
@thermokarst Thanks for the review! Ive updated the code to reflect your suggested changes

jwdebelius:
Thanks @thermokarst. The edits look good to me.

thermokarst:
Sweet, thanks!

thermokarst:
Woot! This is so exciting, thanks again @jwdebelius!

Can you update the [changelog](https://forum.qiime2.org/t/preview-qiime-2-2020-5-development-changelog/13906) when you get a chance?

PR #235, opened 2020-06-18
IMP: adds regression test to catch unusual sort behavior 

### LINKED ISSUE #169: `summarize`: numeric sample IDs cause feature count table to sort incorrectly
**Bug Description**
It looks like we are [sorting the table](https://github.com/qiime2/q2-feature-table/blob/master/q2_feature_table/_summarize/_visualizer.py#L109), but either there is something wrong with `pandas` and numeric IDs, or, the D3 code and numeric IDs.

**References**
1. [sorting the table](https://github.com/qiime2/q2-feature-table/blob/master/q2_feature_table/_summarize/_visualizer.py#L109)
2. This recently [came up on the forum](https://forum.qiime2.org/t/strange-behavior-in-feature-table-summarize/3600).

ChrisKeefe:
closes #169

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ target/
 
 .DS_Store
 node_modules
+
+# .vscode stuff
+.vscode/
--- a/q2_feature_table/tests/test_summarize.py
+++ b/q2_feature_table/tests/test_summarize.py
@@ -300,6 +300,43 @@ class SummarizeTests(TestCase):
                           'SampleType': 'gut'}, 'frequency': 5.0}],
                         spec['data'][0]['values'])
 
+    def test_counts_sort_in_descending_order(self):
+        # attempts to prevent reversions to the unusual sort behavior seen in
+        # https://github.com/qiime2/q2-feature-table/issues/169
+        table = biom.Table(np.array([[0, 1, 3],
+                                    [1, 1, 2],
+                                    [100, 150, 500],
+                                    [1000, 10000, 100000],
+                                    [52, 42, 99]]),
+                           ['O1', 'O2', '0a', 'aa', 'a0'],
+                           ['1', '3', '2'])
+
+        with tempfile.TemporaryDirectory() as output_dir:
+            summarize(output_dir, table)
+
+            feature_freq_fp = os.path.join(output_dir,
+                                           'feature-frequency-detail.csv')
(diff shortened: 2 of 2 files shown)


thermokarst:
Thanks @ChrisKeefe!

## RE Feature IDs:

#169 isn't about feature IDs (as far as I can tell, there appears to be no issue with sorting of feature IDs reported there - the "feature count" table referenced in #169 is the feature count _by sample_ table). If you do want to write a test to check for the sorting, it makes more sense to me to check the sorting in the HTML, not in the CSV. Users can sort CSV at will, but they aren't able to sort the HTML (readily). As well, the sorting issue in #169 is about HTML-sorting (not CSV sorting, but again, different axis - samples vs features).

## RE Sample IDs:

Unfortunately, the test proposed here hits none of the code paths mentioned in #169. As I mentioned in #169 

> I would propose writing up a test or two to prevent regressions, but in this case, I'm not sure if its worth doing (it might not be possible without adding in a browser-render step into our testing pipeline, which is not at all something that is worth our time right now).

> If you want to spend 15 minutes looking at what it would take to add a test, please do

I took a few minutes to investigate, and found code that supports my "hunch" - the sample table sorting is all being done client-side, in the browser:

https://github.com/qiime2/q2-feature-table/blob/1eca3dcb7311d875f45085b6c44c41bcd3c1bb12/q2_feature_table/_summarize/summarize_assets/sample-frequency-detail.html#L181-L199

Unfortunately we don't have the infrastructure to run in-browser tests right now.


## RE Vscode files:

This isn't specific to this PR, but something I have noticed on many of your PRs (including this one) - you include editor-specific gitignore directives - this is something that should preferably be handled by your own global gitconfig on your machine, see mine for example:

https://github.com/thermokarst/dotfiles/tree/master/git

ChrisKeefe:
All true, @thermokarst. I shouldn't have made the assumption that catching unusual behavior in the lines mentioned [here ](https://github.com/qiime2/q2-feature-table/issues/169#issuecomment-632377511) was a good approach. Fun exercise, good learning experience, happy to close. 

Re: .vscode files, thanks for the inspiration. I've been updating per-repo .gitignores as I go because I think most of the lab is using VSCode for development. It's definitely cleaner to handle locally. :1st_place_medal:

Issue #43, opened 2016-10-26
filter-features additional functionality


nbokulich:
I apologize if the following is already planned, or if I have misunderstood what this method is actually doing. It is also possible that some of what I have written below is possible with `where` clauses — in which case I recommended documenting each of these cases when docs are released, as I believe these could be fairly standard uses.

**Input:** this method only accepts `FeatureTable[Frequency]`. It would be great to use on `FeatureTable[RelativeFrequency]`, e.g., for filtering by relative abundance.

**Filtering Criteria:** this method currently only filters on min/max frequency SUMs for a given OTU, which is useful for many situations (e.g., remove all singletons) but not all. A few other cases that would be useful:
1) Relative frequencies: filter by relative frequencies. This is useful for many cases, and both sum abundance and the criteria suggested below should be applicable to either freq or relative freq
2) min/max average freq/relative freq: e.g., remove features that comprise <1% average abundance across samples
3) min/max minimum freq/relative freq: e.g., remove features that are not detected at a minimum of 5% in at least one sample, or remove features detected at more than 5% relative abundance in at least one sample (the latter is probably less useful, but who knows).
4) min/max maximum freq/relative freq: e.g., remove features that are not detected at 5% in at least one sample, or those that are detected above a certain maximum threshold. The former is a very useful method, e.g., for filtering before making taxonomy barplots, so that only taxa that are abundant in at least one sample are shown.

**What to do with filtered features?:** In relative frequency tables, it may be useful to keep relative frequencies intact after filtering, e.g., if a user wants to view abundant taxa on a barplot without altering actual relative frequencies (e.g., filtered groups will appear as empty space or as "Other"), or focus on specific clades without altering relative frequencies. Both of these examples relate to taxonomy tables, so may be more relevant for `q2-taxa` `barplots` instead...

gregcaporaso:
[forum x-ref](https://forum.qiime2.org/t/filter-feature-table-based-on-percent-prevalent/3499)

nbokulich:
[forum xref](https://forum.qiime2.org/t/filter-features-based-on-proportion/4535)

mestaki:
[form xref ](https://forum.qiime2.org/t/relative-abundances-of-taxonomy-analysis/4939/2)

lizgehret:
Closed by #237

PR #237, opened 2020-08-05
Creates conditional feature filtering

jwdebelius:
filters features based on the prevalence of the feature with a relative abundance of at least x. (For example, 0.1% in at least 10% of samples).

Sorry for no previous heads up, or no formal heads up, I was struck by inclination and... this kind of happened

Addresses part of #43


q2_feature_table/plugin_setup.py
```diff
@@ -327,6 +327,34 @@
                 "on https://docs.qiime2.org for additional details."
 )
 
+plugin.methods.register_function(
+    function=q2_feature_table.filter_features_conditionally,
+    inputs={'table': FeatureTable[T1]},
+    parameters={'prevelance': Float % Range(0, 1),
```
andrewsanchez, line 333:
```suggestion
    parameters={'prevalance': Float % Range(0, 1),
```

q2_feature_table/plugin_setup.py
```diff
@@ -327,6 +327,34 @@
...
+        'abundance': ('the minimum relative abundance for a feature to be '
+                      'retained'),
+        'prevelance': ('the minimum portion of samples a that a feature '
+                       'must have a relative abundancee of at least '
+                       '`abundance` to be retained')
+    },
+    output_descriptions={
+        'filtered_table': 'The resulting feature table filtered by feature.'
+    },
+    name="Filter features from a table based on abundance and prevelance",
```
andrewsanchez, line 349:
```suggestion
    name="Filter features from a table based on abundance and prevalence",
```

q2_feature_table/plugin_setup.py
```diff
@@ -327,6 +327,34 @@
...
+    outputs=[('filtered_table', FeatureTable[T1])],
+    input_descriptions={
+        'table': 'The feature table from which features should be filtered.'
+    },
+    parameter_descriptions={
+        'abundance': ('the minimum relative abundance for a feature to be '
+                      'retained'),
+        'prevelance': ('the minimum portion of samples a that a feature '
+                       'must have a relative abundancee of at least '
+                       '`abundance` to be retained')
```
andrewsanchez, line 344:
```suggestion
        'prevalence': ('the minimum portion of samples that a feature '
                       'must have a relative abundance of at least '
                       '`abundance` to be retained')
```

q2_feature_table/plugin_setup.py
```diff
@@ -327,6 +327,34 @@
...
+    output_descriptions={
+        'filtered_table': 'The resulting feature table filtered by feature.'
+    },
+    name="Filter features from a table based on abundance and prevelance",
+    description=("Filter features based on the relative abundance in a "
+                 "certain portion of samples. (i.e. Feature must have a "
+                 "relative abundance of at least `abundance` in at least "
+                 "`prevelance` number of samples). Any samples with a "
+                 "frequency of zero after feature filtering will also be "
+                 "removed.")
```
andrewsanchez, line 355:
```suggestion
    description=("Filter features based on the relative abundance in a "
                 "certain portion of samples (i.e., features must have a "
                 "relative abundance of at least `abundance` in at least "
                 "`prevalence` number of samples). Any samples with a "
                 "frequency of zero after feature filtering will also be "
                 "removed.")
```

q2_feature_table/plugin_setup.py
andrewsanchez: Looks like the spelling error occurs elsewhere in the code as well. ✏️  🤓

q2_feature_table/plugin_setup.py
andrewsanchez: Another minor point here is to begin docstrings with capital letters.

q2_feature_table/_conditional_filter.py
```diff
@@ -0,0 +1,33 @@
...
+    num_observations, num_samples = table.shape
+    prevelance = prevelance * num_samples
+
+    # Calculates the filteering parameters on the original table
+    def _filter_f(v, id_, md):
+        return (v > abundance).sum() >= prevelance
+
+    # Normalized the table to get the prevelance
+    table_n = table.copy().norm(axis='sample', inplace=False)
+    table_s = table_n.filter(_filter_f, axis='observation', inplace=True)
```
andrewsanchez, line 28: Since you _are_ modifying this table in place, assigning to a new variable isn't necessary.  Is there a reason you prefer assigning to a new variable?

I would also suggest using more descriptive variable names, e.g., `table_normalized`.

q2_feature_table/_conditional_filter.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    num_observations, num_samples = table.shape
+    prevelance = prevelance * num_samples
+
+    # Calculates the filteering parameters on the original table
+    def _filter_f(v, id_, md):
+        return (v > abundance).sum() >= prevelance
+
+    # Normalized the table to get the prevelance
+    table_n = table.copy().norm(axis='sample', inplace=False)
```
andrewsanchez, line 27: Since you are not normalizing the table in place, I believe the call to `.copy()` can be dropped.

q2_feature_table/_conditional_filter.py
```diff
@@ -0,0 +1,33 @@
...
+                                  ) -> biom.Table:
+    """
+    A function to perform joint filtering because it makes life better
+    """
+    num_observations, num_samples = table.shape
+    prevelance = prevelance * num_samples
+
+    # Calculates the filteering parameters on the original table
+    def _filter_f(v, id_, md):
+        return (v > abundance).sum() >= prevelance
```
andrewsanchez, line 24: Same suggestion as below:  since we have plenty of space here, I suggest more verbose naming, e.g., `values`, and `metadata`.

andrewsanchez:
Thanks, @jwdebelius and sorry it's taken me a while to get to this!  Looking forward to working on this one!  😄 

My review is incomplete at this point; I added just a few minor inline suggestions/comments about the docstrings.

Before we go too much further with this, I wanted to get @thermokarst's take on API design and code structure.  

Firstly, #43 seemed to suggest that this functionality could be added to the `filter-features` action, whereas this PR creates a new action, `filter-features-conditionally`.  Do you have an opinion regarding extending the functionality of the `filter-features` action versus creating a new action?  I'm not even sure if that would make sense or if it's feasible, I'm just going off of my impression from #43.

Secondly, it seems that `filter_features_conditionally` might be better placed in the existing `_filter.py` file rather than the new `_conditional_filter.py`.

These may be fairly minor concerns, but I am curious nonetheless.

q2_feature_table/_conditional_filter.py
jwdebelius: Biom normalization is tenacious. I run this function probably 3 times a month (I think the PR is copied directly from a gist I've got on my computer but not github). Through trial and error, I've found that if you copy before you normalize you insure your original table remains the same. If you don't, there's a non-zero probability biom will try to be friendly and helpful normalize your data for you anyway because you must have been wrong. `table.copy().norm(inplace=False)` makes sure that you keep your original table intact, which is one of the key pieces to this function. 

I can add some more documentation if that would help clear up the confusion, but the `.copy()` here is intentional and crucial.

jwdebelius:
Thanks @andrewsanchez! I'll go through and address the spelling mistakes. Thanks for proof reading for me! I promise all future pull requests will contain at least 10% more spelling errors than your average, even after running through google docs. 

I had initially planned to put it in the `filter_features` function. I think for me the confusion there is adding an additional argument for the conditionals. Since conditional `abundance` and `prevalence` values are bounded floats between 0 and 1 and interdependent, while the `min_frequency`/`max_frequency` and `min_samples`/`max_samples` are unbounded integers, does it get confusing for users? (I still find the frequency language confusing TBH because to me "frequency" = relative abundance and "counts" = absolute abundance. For me, abundance represents the default quantity 😄 ). So, I think mixing the language will potentially lead to more, rather than less confusion.

q2_feature_table/plugin_setup.py
jwdebelius: Gotta love pep8

q2_feature_table/_conditional_filter.py
jwdebelius: This is the syntax used in the [biom documentation](https://biom-format.org/documentation/generated/biom.table.Table.filter.html#biom.table.Table.filter). When given the choice, I like to be consistent with source anonymous function documentation, I think it makes it easier for people working between this code, the biom example code, and their own code to follow what, where, and how the arguments translate in the different cases

q2_feature_table/_conditional_filter.py
andrewsanchez: That's interesting!  Thanks for clarifying. I wonder why that is.  Nothing in the BIOM code jumps out to me. After a cursory search, the only thing I found where I could see this happening is if you made a call to `Table.collapse` w/o specifying `norm=False`.  Is this an issue you've seen discussed by others anywhere?  If we can reproduce the issue reliably, it should be pretty trivial to expand biom's test suite to cover this edge case you've observed (PR which implemented `inplace` is [here](https://github.com/biocore/biom-format/pull/351/files)).  I have a TON to learn in this realm, FYI.  😃    I'll say more about documentation later.

thermokarst:
> Do you have an opinion regarding extending the functionality of the `filter-features` action versus creating a new action?

I would _prefer_ to see it all in one method, but if @jwdebelius thinks the user interface might be too confusing, I'll support a separate command. One question for @jwdebelius - do you have any suggestions for alternative names for this function? I'm not sure yet if we should think about setting a different name for it or not, I am just looking for some synonyms that might help make the use of "conditionally" _click_ for me more clearly.

> Secondly, it seems that `filter_features_conditionally` might be better placed in the existing `_filter.py` file rather than the new `_conditional_filter.py`

I agree, this is fairly minor, but would be nice. An alternative (and this can be addressed in a subsequent PR if we decide to do it) is to turn `_filter.py` into a subpackage - then the subpackage could have all kinds of `_filter_foo.py` files.

q2_feature_table/_conditional_filter.py
andrewsanchez: That's a really good point.  It still seems somewhat ambiguous to me though since the docstring for `ids_to_keep` uses the verbose var names `values`, and `metadata` and the examples don't use the `v` shorthand (they use `val`).  I'm not really sure what the best practice is here (if there is one), to be honest.  My inclination is to prioritize readability and ergonomics in our own codebase.  I'm not sure how much we gain by staying consistent with biom's examples; and those examples could change.  I'm happy to defer to your judgment here since you have more experience.  Maybe we can have the best of both worlds with good documentation. 🤔 

Another nitpick:  `md` to `_` since it isn't actually used. But again, we can make an exception in this case if there is value in keeping the function signature consistent with the biom doc examples.

jwdebelius: I can switch to the more verbose documentation, and I'd strongly like to keep the metadata argument. I (and I assume other people) learn a lot about the way code is put together by reading codebases. I know pep8/flake8 is supposed to make things more elegant. And I think there's something useful about knowing `biom.filter` requires a 3 argument function (`f(val, id_, md)`) and collapse is a two argument function (`f(id_, md)`) by seeing it in place?

jwdebelius:
> One question for @jwdebelius - do you have any suggestions for alternative names for this function? I'm not sure yet if we should think about setting a different name for it or not, I am just looking for some synonyms that might help make the use of "conditionally" click for me more clearly.

One thought is "joint" filtering, because we're filtering prevalence and abundance together.  (Conditional to me was maybe baysian: x | y).

> I agree, this is fairly minor, but would be nice. An alternative (and this can be addressed in a subsequent PR if we decide to do it) is to turn _filter.py into a subpackage - then the subpackage could have all kinds of _filter_foo.py files.

I moved it over 🤷‍♀️

q2_feature_table/_filter.py
```diff
@@ -111,3 +111,29 @@ def filter_seqs(data: pd.Series, table: biom.Table = None,
...
+                                  ) -> biom.Table:
+    """
+    A function to perform joint filtering because it makes life better
+    """
+    num_observations, num_samples = table.shape
+    prevalence = prevalence * num_samples
+
+    # Calculates the filtering parameters on the original table
+    def _filter_f(value, id_, metadata):
+        return (value >= abundance).sum() >= prevalence
```
andrewsanchez, line 127:
```suggestion
    def _filter_f(values, id_, metadata):
        return (values >= abundance).sum() >= prevalence
```

q2_feature_table/plugin_setup.py
```diff
@@ -327,6 +327,36 @@
...
+        'filtered_table': 'The resulting feature table filtered by feature.'
+    },
+    name="Filter features from a table based on abundance and prevalence",
+    description=("Filter features based on the relative abundance in a "
+                 "certain portion of samples (i.e., features must have a "
+                 "relative abundance of at least `abundance` in at least "
+                 "`prevalence` number of samples). Any samples with a "
+                 "frequency of zero after feature filtering will also be "
+                 "removed.")
+)
```
andrewsanchez, line 357: I think it would be more user friendly if the order of the parameters (and consequently, the order in which they are presented to the user) matched up with the order in which they are discussed in the description.  For example, if we register the `abundance` parameter first, we would get:

```Usage: qiime feature-table filter-features-conditionally [OPTIONS]

  Filter features based on the relative abundance in a certain portion of
  samples (i.e., features must have a relative abundance of at least
  `abundance` in at least `prevalence` number of samples). Any samples with
  a frequency of zero after feature filtering will also be removed.

Inputs:
  --i-table ARTIFACT FeatureTable[Frequency¹ | RelativeFrequency² |
    PresenceAbsence³ | Composition⁴]
                       The feature table from which features should be
                       filtered.                                    [required]
Parameters:
  --p-abundance PROPORTION
    Range(0, 1)        The minimum relative abundance for a feature to be
                       retained.                                    [required]
  --p-prevalence PROPORTION
    Range(0, 1)        The minimum portion of samples that a feature must
                       have a relative abundance of at least `abundance` to be
                       retained.                                    [required]
```
instead of this:

```
Usage: qiime feature-table filter-features-conditionally [OPTIONS]

  Filter features based on the relative abundance in a certain portion of
  samples (i.e., features must have a relative abundance of at least
  `abundance` in at least `prevalence` number of samples). Any samples with
  a frequency of zero after feature filtering will also be removed.

Inputs:
  --i-table ARTIFACT FeatureTable[Frequency¹ | RelativeFrequency² |
    PresenceAbsence³ | Composition⁴]
                       The feature table from which features should be
                       filtered.                                    [required]
Parameters:
  --p-prevalence PROPORTION
    Range(0, 1)        The minimum portion of samples that a feature must
                       have a relative abundance of at least `abundance` to be
                       retained.                                    [required]
  --p-abundance PROPORTION
    Range(0, 1)        The minimum relative abundance for a feature to be
                       retained.                                    [required]
```

The easiest solution would be to modify the description to match up with the current ordering.  Alternatively, we can change the order of the parameters wherever they appear to match up with the docstring.  Whichever solution you prefer is fine with me!  What do you think?

andrewsanchez (changes requested):
Hi, @jwdebelius!  Thanks again for your patience with this 😃 . This is looking great and almost ready to be merged in! 🚀 

I added two inline comments.  If you'd like, I'd be happy to push a commit with my suggestions to your PR branch for you to review.

jwdebelius:
Hi @andrewsanchez, 

I *think* I addressed you comments. Thanks for the review!

andrewsanchez (approved):
Fabulous!  Looks great.  Thanks, @jwdebelius!

PR #242, opened 2021-01-19
ENH: added average and RelativeFrequency handling to merge

BenKaehler:
Added an `average` `overlap_method` to `merge`.

Enabled `merge` to process `RelativeFrequency` `FeatureTable`s.

### DIFF
--- a/q2_feature_table/_merge.py
+++ b/q2_feature_table/_merge.py
@@ -13,7 +13,7 @@ import collections
 
 def overlap_methods():
     return ('error_on_overlapping_sample', 'error_on_overlapping_feature',
-            'sum')
+            'sum', 'average')
 
 
 def _get_overlapping(tables, axis):
@@ -42,11 +42,14 @@ def merge(tables: biom.Table,
             overlapping = _get_overlapping(tables, 'observation')
             raise ValueError('Same features are present in some of the '
                              'provided tables: %s' % ', '.join(overlapping))
-    elif overlap_method == 'sum':
+    elif overlap_method in ('sum', 'average'):
+        n = len(tables)
         tables = iter(tables)
         result = next(tables)  # There is always at least 1
         for table in tables:
             result = result.merge(table)
+        if overlap_method == 'average':
+            result.transform(lambda v, _, __: v / n)
         return result
     else:
         raise ValueError('Invalid overlap method: %s. Please provide one of '
--- a/q2_feature_table/_merge.py
+++ b/q2_feature_table/_merge.py
@@ -13,7 +13,7 @@ import collections
 
 def overlap_methods():
     return ('error_on_overlapping_sample', 'error_on_overlapping_feature',
-            'sum', 'average')
+            'average', 'sum')
 
 
 def _get_overlapping(tables, axis):
--- a/q2_feature_table/_merge.py
+++ b/q2_feature_table/_merge.py
@@ -12,8 +12,8 @@ import collections
 
 
 def overlap_methods():
-    return ('error_on_overlapping_sample', 'error_on_overlapping_feature',
-            'average', 'sum')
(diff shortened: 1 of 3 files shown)


andrewsanchez (commented):
Just took a first pass on this and it looks good to me.  Thanks, @BenKaehler!  @ebolyen, @thermokarst mentioned you might have some thoughts about naming here.  I think he mentioned average vs. mean, in particular.

q2_feature_table/tests/test_merge.py
```diff
@@ -167,6 +167,16 @@ def test_sum_overlapping_feature_ids(self):
                     ['S1', 'S2', 'S3', 'S4', 'S5', 'S6'])
         self.assertEqual(obs, exp)
 
+    def test_average(self):
```
thermokarst, line 170: @BenKaehler, do you think we should add a test for a relative frequency table, too?

BenKaehler: Erm, at that level they’re all biom Tables. Should we build another test that calls it via the plugin interface?

thermokarst: Could you just make a vanilla biom table with fractional values?

thermokarst: I added a quick test for this.

q2_feature_table/plugin_setup.py
```diff
@@ -174,14 +174,25 @@
...
 )
 
+i_table, p_overlap_method, o_table = TypeMap({
+    (FeatureTable[Frequency],
+     Str % Choices(sorted(q2_feature_table.overlap_methods()))):
+    FeatureTable[Frequency],
+    (FeatureTable[RelativeFrequency],
+     # We don't want to allow summing of RelativeFrequency tables, so remove
+     # that option from the overlap methods
+     Str % Choices(sorted(q2_feature_table.overlap_methods() - {'sum'}))):
```
thermokarst, line 184: I made a minor tweak here by converting the `overlap_methods` to a set, that way we can clearly say which element is being removed here, rather than doing some indexing arithmetic in my head.

q2_feature_table/_merge.py
```diff
@@ -25,7 +25,8 @@ def _get_overlapping(tables, axis):
 
 def merge(tables: biom.Table,
           overlap_method: str = 'error_on_overlapping_sample') -> biom.Table:
-    if len(tables) == 1:
+    n_tables = len(tables)
```
thermokarst, line 28: I figured no need to compute the length twice - and also I think it helps stitch it all together (I didn't even notice we were computing twice until I expanded the diff).

thermokarst (approved):
This LGTM, thanks @BenKaehler! Please let me know if my edits are acceptable - thanks!

q2_feature_table/plugin_setup.py
BenKaehler: LGTM, thanks @thermokarst

q2_feature_table/_merge.py
BenKaehler: That’s [fine](https://wiki.c2.com/?PrematureOptimization) :).

Thanks again.

BenKaehler:
> This LGTM, thanks @BenKaehler! Please let me know if my edits are acceptable - thanks!

The tweaks are more than acceptable. I am happy with this version. Thanks very much @thermokarst and @andrewsanchez!

q2_feature_table/_merge.py
thermokarst: Hah! I don't know the first thing about optimization, let alone premature optimization.

thermokarst:
@BenKaehler   - do you think you can add a short entry for this change into the release changelog?

https://forum.qiime2.org/t/preview-qiime-2-2021-2-development-changelog/17815

PR #243, opened 2021-02-11
Added keep_zero_features option to filter-samples which addresses iss…

### LINKED ISSUE #185: `filter-samples` should not drop empty features by default
**Current Behavior**
After filtering samples, "Any features with a frequency of zero after sample filtering will also be removed." [according to the docs](https://docs.qiime2.org/2018.6/plugins/available/feature-table/filter-samples/)

Silently dropping these features might be undesirable to some users.

**Proposed Behavior**
Should add a parameter to choose whether or not empty features should be dropped, or even a threshold (at that point making a pipeline that filters samples and features might be a better solution). Otherwise, don't drop empty features. 

Issuing a warning even just to stdout could also be useful, indicating the # and IDs of features that were dropped.

**References**
1. [according to the docs](https://docs.qiime2.org/2018.6/plugins/available/feature-table/filter-samples/)
2. [forum xref](https://forum.qiime2.org/t/feature-table-filter-samples/5323)

dwthomas:
fixes #185 
Added keep-zero-features as an option, to filter-samples.  Default behavior is maintained, but if keep-zero-features is true then features with zero frequency in the filtered table are not dropped.

### DIFF
--- a/q2_feature_table/_filter.py
+++ b/q2_feature_table/_filter.py
@@ -32,7 +32,7 @@ _other_axis_map = {'sample': 'observation', 'observation': 'sample'}
 
 def _filter_table(table, min_frequency, max_frequency, min_nonzero,
                   max_nonzero, metadata, where, axis, exclude_ids=False,
-                  keep_zero_features=False):
+                  filter_opposite_axis=True):
     if min_frequency == 0 and max_frequency is None and min_nonzero == 0 and\
        max_nonzero is None and metadata is None and where is None and\
        exclude_ids is False:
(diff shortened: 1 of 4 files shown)


thermokarst (approved):
LGTM, thanks @dwthomas! I made a few minor tweaks to this, including tweaking the parameter name, exposing the same behavior on both axes filtering methods, and adding some basic unit tests.

thermokarst:
@dwthomas   - do you think you can add a short entry for this change into the release changelog?

https://forum.qiime2.org/t/preview-qiime-2-2021-2-development-changelog/17815

PR #264, opened 2022-10-24
ex: adds usage examples for Hacktoberfest

Oddant1:
Add usage examples for merge-seqs, filter-samples, filter-features, and filter-features-conditionally.


Oddant1:
In regards to https://github.com/qiime2/q2-feature-table/pull/264/commits/5a9683ba35f56421fa3cc9426334f3144fc5d61d, I was only able to find one FeatureData[Taxonomy] artifact in the [cancer microbiome tutorial](https://docs.qiime2.org/jupyterbooks/cancer-microbiome-intervention-tutorial/030-tutorial-downstream/020-taxonomy.html), so the example is currently merging a table with itself. It would probably be better to have two unique artifacts if anyone knows where to get another from the tutorials, or if it would be worthwhile to make our own little ones for this example.

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,357 @@
...
+
+def ft3_factory():
+    return Artifact.import_data(
+        'FeatureTable[Frequency]',
+        Table(np.array([[0, 4, 9], [4, 4, 8]]),
+              ['O1', 'O4'],
+              ['S7', 'S8', 'S9']))
+
+
+def feature_table_merge_example(use):
```
lizgehret, line 53:
```suggestion
def feature_table_merge_two_tables(use):
```
I would change this test name to just the method/use case where you are running (since all of these tests are within the `_examples.py` file already.

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,357 @@
...
+
+    merged_table, = use.action(
+        use.UsageAction(plugin_id='feature_table',
+                        action_id='merge'),
+        use.UsageInputs(tables=[feature_table1, feature_table2]),
+        use.UsageOutputNames(merged_table='merged_table'),
+    )
+
+
+def feature_table_merge_three_tables_example(use):
```
lizgehret, line 65:
```suggestion
def feature_table_merge_three_tables(use):
```
Same here, I'd just remove example from the test name.

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,357 @@
...
+        use.UsageInputs(
+            data=[seqs1, seqs2]
+        ),
+        use.UsageOutputNames(
+            merged_data='merged_data'
+        )
+    )
+
+
+def feature_table_filter_seqs_metadata1(use):
```
lizgehret, line 97:
```suggestion
def feature_table_filter_seqs_subject1(use):
```
I'd also change this name (and the following test as well) to indicate the different ways you're filtering. I was initially wondering if you were using two different metadata files for your filtering.

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,357 @@
...
+
+    filtered_seqs, = use.action(
+        use.UsageAction(plugin_id='feature_table', action_id='filter_seqs'),
+        use.UsageInputs(table=seqs, metadata=sample_metadata,
+                        where='[subject]="subject-1"'),
+        use.UsageOutputNames(filtered_seqs='filtered_seqs')
+    )
+
+
+def feature_table_filter_seqs_metadata2(use):
```
lizgehret, line 113:
```suggestion
def feature_table_filter_seqs_bodysite_leftpalm_rightpalm(use):
```
Same as above comment - not sure if this is the best name change, but probably good to add something that indicates the way you're filtering here.

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,357 @@
...
+
+    filtered_seqs, = use.action(
+        use.UsageAction(plugin_id='feature_table', action_id='filter_seqs'),
+        use.UsageInputs(table=feature_table, metadata=sample_metadata,
+                        where='[body-site] IN ("left palm", "right palm")'),
+        use.UsageOutputNames(filtered_seqs='filtered_seqs')
+    )
+
+
+def feature_table_filter_seqs_metadata3(use):
```
lizgehret, line 129:
```suggestion
def feature_table_filter_seqs_subject_and_bodysite(use):
```
Same as above comments re: naming convention.

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,357 @@
...
+
+    filtered_seqs, = use.action(
+        use.UsageAction(plugin_id='feature_table', action_id='filter_seqs'),
+        use.UsageInputs(table=feature_table, metadata=sample_metadata,
+                        where=r'[subject]="subject-1" AND [body-site]="gut"'),
+        use.UsageOutputNames(filtered_seqs='filtered_seqs')
+    )
+
+
+def feature_table_filter_seqs_metadata4(use):
```
lizgehret, line 145: Maybe a better way to adjust the naming conventions for these different filter_seqs tests is to add the filtering operator that you're testing - so something like:
```suggestion
def feature_table_filter_seqs_OR(use):
```
I'm not attached one way or the other, it was just another thought I had.

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,357 @@
...
+    filtered_seqs, = use.action(
+        use.UsageAction(plugin_id='feature_table', action_id='filter_seqs'),
+        use.UsageInputs(
+            table=feature_table, metadata=sample_metadata,
+            where=r'[body-site]="gut" OR [reported-antibiotic-usage]="Yes"'),
+        use.UsageOutputNames(filtered_seqs='filtered_seqs')
+    )
+
+
+def feature_table_filter_seqs_metadata5(use):
```
lizgehret, line 162:
```suggestion
def feature_table_filter_seqs_AND_NOT(use):
```
🤷‍♀️

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,357 @@
...
+        use.UsageAction(plugin_id='feature_table', action_id='filter_seqs'),
+        use.UsageInputs(
+            table=feature_table, metadata=sample_metadata,
+            where=r'[subject]="subject-1" AND NOT [body-site]="gut"'),
+        use.UsageOutputNames(filtered_seqs='filtered_seqs')
+    )
+
+
+def feature_table_merge_taxa(use):
+    # TODO: Would probably be better to have two different artifacts here
```
lizgehret, line 180: Hmm yeah, I agree. For the purposes of adding this example, I suppose it's fine for now - but it would definitely be more functional from a testing perspective if we had two different taxa artifacts.

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,357 @@
...
+
+    filtered_table, = use.action(
+        use.UsageAction(plugin_id='feature_table', action_id='filter_samples'),
+        use.UsageInputs(table=feature_table,
+                        min_frequency=1500),
+        use.UsageOutputNames(filtered_table='filtered_table')
+    )
+
+
+def feature_table_filter_samples_metadata1(use):
```
lizgehret, line 222: Same thoughts for these subsequent examples as with the `filter_seqs` examples - I'd modify the names to be more descriptive as to how you're filtering.

q2_feature_table/plugin_setup.py
```diff
@@ -338,7 +353,15 @@
     description="Filter samples from table based on frequency and/or "
                 "metadata. Any features with a frequency of zero after sample "
                 "filtering will also be removed. See the filtering tutorial "
-                "on https://docs.qiime2.org for additional details."
+                "on https://docs.qiime2.org for additional details.",
+    examples={
+        'filter_to_subject1': feature_table_filter_samples_metadata1,
+        'filter_to_skin': feature_table_filter_samples_metadata2,
+        'filter_to_subject1_gut': feature_table_filter_samples_metadata3,
+        'filter_to_gut_or_abx': feature_table_filter_samples_metadata4,
```
lizgehret, line 361: These are good user-facing names - maybe just match the example names with the user facing naming conventions? Sorry, I've suggested like three different ways to change the names lol - I think this might be the best approach though.

q2_feature_table/plugin_setup.py
```diff
@@ -367,7 +390,11 @@
                  "relative abundance of at least `abundance` in at least "
                  "`prevalence` number of samples). Any samples with a "
                  "frequency of zero after feature filtering will also be "
-                 "removed.")
+                 "removed."),
+    examples={
+        '1':
```
lizgehret, line 395: I'd adjust this to just match the method name.

lizgehret (changes requested):
Overall this lgtm, thanks for adding really thorough examples @Oddant1! Just a few comments inline regarding naming conventions.

Oddant1:
It turns out I was silly and wasn't running the merge_taxa tests, and they didn't work when I did run them anyway lol. I removed them and changed the names of the other examples.

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,278 @@
...
+rep_seqs_2_url = (f'https://data.qiime2.org/{qiime2.__release__}/'
+                  'tutorials/phylogeny/rep-seqs.qza')
+taxonomy_1_url = ('https://docs.qiime2.org/jupyterbooks/cancer-microbiome-'
+                  'intervention-tutorial/data/030-tutorial-downstream/020-'
+                  'taxonomy/taxonomy.qza')
+moving_pics_ft_url = (f'https://data.qiime2.org/{qiime2.__release__}/'
+                      'tutorials/filtering/table.qza')
+moving_pics_md_url = (f'https://data.qiime2.org/{qiime2.__release__}/'
+                      'tutorials/moving-pictures/sample_metadata.tsv')
+rep_seqs_dada2_url = 'https://docs.qiime2.org/2022.8/data/tutorials/' \
```
lizgehret, line 27: Not sure if we should hard-code the QIIME 2 release in these URLs... I would try using the qiime2.__release__ for these rep seqs URLs and see if they work!

lizgehret (commented):
All of these changes look good outside of my one comment about those URLs. I'll pull this down locally to do a quick test, but this should be basically ready to merge!

q2_feature_table/_examples.py
Oddant1: Whoops yeah those are from an older commit and didn't get updated good catch

lizgehret, line 27:
```suggestion
rep_seqs_dada2_url = 'https://data.qiime2.org/usage-examples/' \
```

q2_feature_table/_examples.py
```diff
@@ -0,0 +1,278 @@
...
+                     'moving-pictures/rep-seqs-dada2.qza'
+rep_seqs_deblur_url = 'https://docs.qiime2.org/2022.8/data/tutorials/' \
```
lizgehret, line 29:
```suggestion
rep_seqs_deblur_url = 'https://data.qiime2.org/usage-examples/' \
```

lizgehret (approved):
just updated the relevant URLs with usage links from data302, so once those changes pass on CI this is good to merge!

PR #265, opened 2022-12-02
Rarefy with replacement bug

wasade:
Rarefy `with_replacement=True` was incorrectly retaining samples with depth below the requested subsampling level. This behavior differed from `with_replacement=False`, where samples below depth are removed from the table. This Pull Request corrects the behavior of `with_replacement=True` by removing samples below the requested depth prior to subsampling.

### DIFF
--- a/q2_feature_table/_normalize.py
+++ b/q2_feature_table/_normalize.py
@@ -11,6 +11,9 @@ import biom
 
 def rarefy(table: biom.Table, sampling_depth: int,
            with_replacement: bool = False) -> biom.Table:
+    if with_replacement:
+        table = table.filter(lambda v, i, m: v.sum() >= sampling_depth,
+                             inplace=False, axis='sample')
     table = table.subsample(sampling_depth, axis='sample', by_id=False,
                             with_replacement=with_replacement)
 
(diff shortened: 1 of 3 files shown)


wasade:
cc @ElDeveloper regarding qiime2/q2-diversity#325

ebolyen:
Thanks @wasade! 

@gregcaporaso and @wasade, do you think this warrants a announcement about the behavior or some special notice in our release notes?

wasade:
That's a good question. I think a special notice in the release notes along
the lines of "Sampling with replacement in rarefaction is now consistent in
handing of samples below depth with its without replacement counterpart.
Importantly, prior versions of QIIME 2 did not explicitly remove samples
below depth if with replacement was specified." The special notice is I
think justified given the recent announcement of support for with
replacement in core diversity.

ElDeveloper (approved):
Thanks for addressing this @wasade!

gregcaporaso:
>  I think a special notice in the release notes along the lines of ...

That sounds like a good plan to me. Thanks for addressing this @wasade!

ebolyen:
Thanks @wasade!

ElDeveloper:
Thanks everyone!

On Dec 7, 2022, at 1:42 PM, Evan Bolyen ***@***.***> wrote:



Thanks @wasade<https://github.com/wasade>!

PR #268, opened 2022-12-16
ex: adds summarize/tabulate-seqs & some misc cleanup

lizgehret:
This PR adds usage examples for the following methods within q2-feature-table:
```
summarize
tabulate-seqs
```
In addition, I've done some minor clean-up to the existing examples to keep formatting consistent and reduce the number of separate imports happening within `plugin_setup.py`.


ebolyen (approved):
Thanks @lizgehret!

PR #277, opened 2023-06-29
NEW: adds feature-table split method

### LINKED ISSUE #275: add split-feature-table action
Add new action, `split-feature-table`, which takes a `FeatureTable` and categorical `MetadataColumn` as input, and outputs a collection of feature tables where the samples in each table are defined by their value in the metadata column. 

For example, if the metadata column looks like the following:

| id | category | 
| --|-- |
| a | 1 |
| b | 2 |
| c | 1 |
| d | 3 |
| e | 2 |
| f | |
| g | 2 |

Three feature tables would be present in the output collection: the first (call it table `1`, corresponding to the category value `1`) would contain samples `a` and `c`; the second (`2`) would contain samples `b`, `e`, and `g`; and the third (`3`) would contain sample `d`. Sample `f` doesn't have a `category` value, so it wouldn't be included in any of the feature tables. 

Two primary motivations for this is that it saves having to run multiple `filter-samples` commands if you want these feature tables, and it will support an iterative `Pipeline` for running `qiime quality-control decontam-identify`. This is functionality that was fairly widely used in QIIME 1, but which we never got around to implementing in QIIME 2.

gregcaporaso:
resolves #275

### DIFF
--- a/q2_feature_table/_split.py
+++ b/q2_feature_table/_split.py
@@ -1,0 +1,28 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016-2023, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import biom
+import qiime2
+
+
+def split(table: biom.Table,
+          metadata: qiime2.CategoricalMetadataColumn,
+          filter_empty_features: bool = True) -> biom.Table:
+
+    metadata_df = metadata.drop_missing_values().to_dataframe()
+
+    indices = metadata_df.reset_index(
+        ).groupby(metadata.name)[metadata_df.index.name].apply(list).to_dict()
+
+    result = {}
+    for group, sample_ids in indices.items():
+        t = table.filter(sample_ids, axis='sample', inplace=False)
+        if filter_empty_features:
+            t.remove_empty(axis='observation', inplace=True)
+        result[group] = t
+    return result
--- a/q2_feature_table/_split.py
+++ b/q2_feature_table/_split.py
@@ -10,28 +10,6 @@ import biom
 import qiime2
 
 
-_valid_value_characters = set(
-    "0123456789"
-    "abcdefghijklmnopqrstuvwxyz"
-    "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
-    "-_.")
-
-
-def _check_valid_value(value):
-    return len(set(value) - _valid_value_characters) != 0
-
-
-def _check_valid_values(values):
-    invalid_values = [v for v in values if _check_valid_value(v)]
-    if len(invalid_values) > 0:
-        raise ValueError(
-            "Invalid value(s) identified during feature table splitting. "
-            "Values in metadata columns used for splitting can only contain "
-            "alphanumeric characters, dashes (-), dots (.), and underscores "
-            "(_). The problematic value(s) are: "
-            f"{', '.join(invalid_values)}")
-
-
 def split(table: biom.Table,
           metadata: qiime2.CategoricalMetadataColumn,
           filter_empty_features: bool = True) -> biom.Table:
@@ -40,7 +18,14 @@ def split(table: biom.Table,
     indices = metadata_df.reset_index(
         ).groupby(metadata.name)[metadata_df.index.name].apply(list).to_dict()
 
-    _check_valid_values(indices.keys())
+    try:
+        qiime2.sdk.util.validate_result_collection_keys(*indices.keys())
+    except KeyError as e:
+        raise KeyError(
+            "One or more invalid metadata column values identified during "
+            "feature table splitting. All metadata column values must be "
+            "valid ResultCollection keys when used for splitting a feature "
+            f"table. The original error message is as follows:\n{str(e)}")
 
     result = {}
     for group, sample_ids in indices.items():
--- a/q2_feature_table/_split.py
+++ b/q2_feature_table/_split.py
@@ -25,7 +25,7 @@ def split(table: biom.Table,
             "One or more invalid metadata column values identified during "
             "feature table splitting. All metadata column values must be "
             "valid ResultCollection keys when used for splitting a feature "
-            f"table. The original error message is as follows:\n{str(e)}")
+            f"table. The original error message is as follows: {str(e)}")
 
     result = {}
     for group, sample_ids in indices.items():
--- a/q2_feature_table/_split.py
+++ b/q2_feature_table/_split.py
@@ -10,14 +10,37 @@ import biom
 import qiime2
 
 
+_valid_value_characters = set(
+    "0123456789"
+    "abcdefghijklmnopqrstuvwxyz"
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+    "-_.")
+
+
+def _check_valid_value(value):
+    return len(set(value) - _valid_value_characters) != 0
+
+
+def _check_valid_values(values):
+    invalid_values = [v for v in values if _check_valid_value(v)]
+    if len(invalid_values) > 0:
+        raise ValueError(
+            "Invalid value(s) identified during feature table splitting. "
+            "Values in metadata columns used for splitting can only contain "
+            "alphanumeric characters, dashes (-), dots (.), and underscores "
+            "(_). The problematic value(s) are: "
+            f"{', '.join(invalid_values)}")
+
+
 def split(table: biom.Table,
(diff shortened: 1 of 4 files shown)


q2_feature_table/_split.py
```diff
@@ -0,0 +1,28 @@
...
+
+import biom
+import qiime2
+
+
+def split(table: biom.Table,
+          metadata: qiime2.CategoricalMetadataColumn,
+          filter_empty_features: bool = True) -> biom.Table:
+
+    metadata_df = metadata.drop_missing_values().to_dataframe()
```
lizgehret, line 17: So this line of code will drop any values within the given md column that don't have a value - should this be distinctly different from the intended functionality of the `filter_empty_features` parameter? The way I'm understanding that parameter, it should do the same thing (if enabled). I tested this out locally and received the same results when this flag was both enabled and disabled, using the following data:
```
sample-id    body-site
L1S8             NaN
L1S57            NaN
L1S76            NaN
L1S105           NaN
L2S155     left palm
L2S175     left palm
L2S204     left palm
L2S222     left palm
L3S242    right palm
L3S294    right palm
L3S313    right palm
L3S341    right palm
L3S360    right palm
L5S104        tongue
L5S155        tongue
L5S174        tongue
L5S203        tongue
L5S222        tongue
L1S140           NaN
L1S208           NaN
L1S257           NaN
L1S281           NaN
L2S240     left palm
L2S309     left palm
L2S357     left palm
L2S382     left palm
L3S378    right palm
L4S63     right palm
L4S112    right palm
L4S137    right palm
L5S240        tongue
L6S20         tongue
L6S68         tongue
L6S93         tongue
```
What I got back were three tables (for 'left palm', 'right palm', and 'tongue', respectively) - I would expect to get back a fourth table when `filter_empty_features` isn't enabled that would include all of the features without an associated group in this md column, is that what you intended?

q2_feature_table/plugin_setup.py
```diff
@@ -607,3 +607,31 @@
...
+    input_descriptions={
+        'table': 'The table to split.'
+    },
+    parameter_descriptions={
+        'metadata': 'A column defining the groups. Each unique value will '
+                    'define a split feature table.',
+        'filter_empty_features': 'If true, features which are not present in '
+                                 'a split feature table are dropped.',
+    },
+    output_descriptions={
```
lizgehret, line 630: I wonder if it's worth including something in this description that clarifies that this output will be a directory containing all of the split feature tables - I know that this is the standard format for a collection, but I wonder if this might be confusing for people who either try to write these tables to an existing directory, or try to list out filenames for the number of split tables they're expecting.

lizgehret (commented):
Couple of in-line comments, but in general this looks good - this will be a really nice method to have! 🙌

q2_feature_table/_split.py
gregcaporaso: @lizgehret, this is defining the behavior for features that would have a total frequency of zero in a split feature table (see my `test_drop_features_w_zero_count` test case). So, for example, if an ASV is only present in `tongue` samples in the input table and `filter_empty_features == False`, that ASV will be included in the output `left palm` and `right palm` tables but it's total frequency in each will be zero (since it wasn't observed in either of those sample types in the input table). If `filter_empty_features == True`, it would not be included in the `left palm` and `right palm` tables. 

Does that make sense? Let me know if you have suggestions for a better parameter name or documentation to make this more clear.

gregcaporaso:
Thanks for the feedback @lizgehret!

While doing a test just now, I realized that since I'm creating file names based on the metadata values, it's possible to end up with spaces (and all kinds of other crazy characters) in the resulting file names. Think I should do something about that, and if so do you know if we have a typical way of doing that in the framework? I feel like url encoding would be overkill. 

```bash
$ qiime feature-table split --i-table table.qza --m-metadata-file sample-metadata.tsv --m-metadata-column body-site --o-tables split-tables

Saved Collection[FeatureTable[Frequency]] to: split-tables
(q2dev-20235-feature-table) ➜  temp ls split-tables
gut.qza        left palm.qza  right palm.qza tongue.qza
```

ebolyen:
Re: crazy names, I thought collections only supported python identifiers (or at least a slightly relaxed form)

I think the best approach (for now) may be to error if a metadata value is not a valid key for a collection.

cc @Oddant1

q2_feature_table/_split.py
lizgehret: Ahh, yes that makes sense - thanks for clarifying! After re-reading your parameter descriptions, I think this seems reasonably clear.

lizgehret:
> Re: crazy names, I thought collections only supported python identifiers (or at least a slightly relaxed form)
> 
> I think the best approach (for now) may be to error if a metadata value is not a valid key for a collection.

That seems reasonable, I can't think of a better solution at this point.

gregcaporaso:
This suggestion sounds good to me, I'll get it in to this PR asap.

gregcaporaso:
I just pushed [a commit](https://github.com/gregcaporaso/q2-feature-table/commit/c23848820ea16faf270a941c6cb87e30b4ed9899) up, but for some reason it hasn't shown up on the PR yet - not sure if it's me or GitHub (I'll check back later). But...

While working on this, I couldn't come up with a way to describe to a user in an error message what a valid Python identifier is, and couldn't find an accessible URL to reference. So I feel like that shouldn't be the criteria here since it may be confusing to users (and lead to forum questions). Plus, I can imagine values such as subject identifiers starting with numbers and there's no reason we can't support that (please correct me if I'm wrong about that, eg b/c of how `Collections` work), so it feels overly restrictive. I instead defined valid values as only containing alphanumerics, and `-`, `.`, and `_`. 

I'm open to feedback on this - maybe we should push this up to the level of `Collections` and let those define what are valid values? I could then catch the error and re-raise it with action-specific context to help with interpretation. 

cc @Oddant1

lizgehret:
I like the changes that you implemented @gregcaporaso but I do wonder if this should just be pushed up to `Collections` as a whole, since this is something that will almost certainly show up in other actions as well.

Oddant1:
@ebolyen cache keys can only be Python identifiers. We can do that for collection keys as well if you want, but right now it just needs to be a string.

gregcaporaso:
Following discussion of this issue in our engineering meeting we're going to bump this validation up to `Collection`. This PR is now blocked by https://github.com/qiime2/qiime2/issues/699.

q2_feature_table/_split.py
```diff
@@ -0,0 +1,51 @@
...
+        ).groupby(metadata.name)[metadata_df.index.name].apply(list).to_dict()
+
+    _check_valid_values(indices.keys())
+
+    result = {}
+    for group, sample_ids in indices.items():
+        t = table.filter(sample_ids, axis='sample', inplace=False)
+        if filter_empty_features:
+            t.remove_empty(axis='observation', inplace=True)
+        result[group] = t
```
wasade, line 50: I think this could be expressed as the following:

```suggestion
    lookup = {}
    for group, sample_ids in indices.items():
        for sid in sample_ids:
            lookup[sid] = group
            
    def group_f(i, m):
        return lookup[i]
        
    result = dict(table.partition(group_f))
```

...I suspect the construction of `lookup` could be done directly from Pandas as well.

For reference, see [`Table.partition`](https://github.com/biocore/biom-format/blob/941973a6aac3bd2621be8818e9e15ccb9e63bdb2/biom/table.py#L2398-L2446)

gregcaporaso: Thanks @wasade - I forgot about the `Table.partition` method.

gregcaporaso: I tried this approach, but it doesn't handle the case where there is missing metadata for sample(s), in which case we'd not include those sample(s) in any of the resulting tables (i.e., my `SplitTests.test_missing_data_samples_dropped`). Is there a way to handle that with `Table.partition`? 

Since we're getting close to release time we can always circle back to this in future.

wasade: Ah good catch. One possibility is to dump them into an "__unsupported__" bin or some unlikely to exist key, and then remove it from the dict after? I agree that would be a helpful API expansion -- I'll open an issue on BIOM

gregcaporaso: Thanks for creating the issue @wasade! I'm going to hold off on using Table.partition here until that feature is added, but will happily delegate all of this to Table.partition then. I made a couple of additional suggestions on that issue that would move all of the interesting bits of this function to the `Table.partition` method (though those suggestions aren't blockers for using this here, just conveniences).

lizgehret (commented):
this lgtm @gregcaporaso - once [this PR](https://github.com/qiime2/qiime2/pull/700) gets merged, we can re-run ci here and make sure this passes before merging.

Issue #252, opened 2021-09-08
in versions 2020.11 and later sample frequency tables have a random ",0" as first row


mestaki:
To reproduce:
Download the `Frequency per sample detail` .csv from the table summary visualizations in version 2020.11 or later ([Moving pictures table summary](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fdocs.qiime2.org%2F2020.11%2Fdata%2Ftutorials%2Fmoving-pictures%2Ftable.qzv))
top 2 rows:

|                                  | 0     |
| -------------------------------- | ----- |
| 4b5eeb300368260019c1fbc7a3c718fc | 11373 |

Now, download the same table from a previous version, here [from 2020.8](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fdocs.qiime2.org%2F2020.8%2Fdata%2Ftutorials%2Fmoving-pictures%2Ftable.qzv)

top 2 rows:
| 4b5eeb300368260019c1fbc7a3c718fc | 11404 |
| -------------------------------- | ----- |
| fe30ff0f71a38a39cf1717ec2be3a2fc | 8933  |

As you can see with the newer table there is an odd blank value and a zero inserted into the first row. I assumed this is unintentional/bug. Not a big deal, but it did break some codes I had for detecting sample frequency depths so figured others may be experiencing something similar.

PR #278, opened 2023-07-12
BUG: Removes 0 as header value when downloading CSVs

cherman2:
fixes https://github.com/qiime2/q2-feature-table/issues/252

When downloading the Frequency per sample and the Frequency per feature CSVs from the feature-table summarize, a 0 was added as a placeholder column name. This was happening pandas .to_csv function, I set header to false and now pandas does not try to create a placeholder column name when creating those CSVs 

This PR does not have tests because it is an extensive process to test downloads from visualizations and this does not effect downstream workflows

### DIFF
--- a/q2_feature_table/_summarize/_visualizer.py
+++ b/q2_feature_table/_summarize/_visualizer.py
@@ -134,9 +134,11 @@ def summarize(output_dir: str, table: biom.Table,
     sample_frequencies.sort_values(inplace=True, ascending=False)
     feature_frequencies.sort_values(inplace=True, ascending=False)
     sample_frequencies.to_csv(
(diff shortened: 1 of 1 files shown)


q2_feature_table/_summarize/_visualizer.py
```diff
@@ -272,7 +274,6 @@ def _frequencies(table, axis):
 
 def _frequency_summary(table, axis='sample'):
     frequencies = _frequencies(table, axis=axis)
-
```
lizgehret, line 275: I would put this new line back in just to keep the diff a bit cleaner, since it's not a part of this PR's functionality.

lizgehret (commented):
lgtm! one small inline comment but otherwise this one should be good to merge.

PR #281, opened 2023-08-23
IMP: enable merge of FeatureTable[PresenceAbsence]

### LINKED ISSUE #238: IMP: `merge` should accept other feature table types
**Improvement Description**
`merge` should accept multiple feature table types, at least `RelativeFrequency` as well as `Frequency`

**Current Behavior**
`merge` only accepts `List[FeatureTable[Frequency]]`

**Proposed Behavior**
Support `RelativeFrequency` and maybe other types. Use `TypeMap` to map the inputs to the outputs (should be the same type)

**Questions**
1. Is `TypeMap` enough to solve this? It seems like a transformer might need to be added to convert to other types (e.g., relative frequency) when outputting, since right now `merge` just sums the values.

gregcaporaso:
This is needed for some of the shotgun workflows. 

Closes #238.

### DIFF
--- a/q2_feature_table/_merge.py
+++ b/q2_feature_table/_merge.py
@@ -13,7 +13,7 @@ import collections
 
 def overlap_methods():
     return {'error_on_overlapping_sample', 'error_on_overlapping_feature',
-            'average', 'sum'}
+            'average', 'sum', 'union'}
 
 
 def _get_overlapping(tables, axis):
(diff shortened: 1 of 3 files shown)


lizgehret:
This all looks reasonable to me @gregcaporaso - I'm just going to pull this down locally to test before merging!

Issue #276, opened 2023-06-28
add optional inputs to `tabulate-seqs`


gregcaporaso:
To assist with assessing why ASVs aren't being assigned taxonomy (one of many [forum x-refs](https://forum.qiime2.org/t/most-of-its-asvs-map-to-other-60/26923/2)), it would be useful if we could optionally provide `FeatureData[Taxonomy]` ~~(and maybe also `FeatureTable[Frequency]`)~~ to the `tabulate-seqs` visualizer. This would add the taxonomy as a new column in the table that is generated by `tabulate-seqs`, making it easier to identify the sequences that are getting poor taxonomy assignments. ~~If the `FeatureTable[Frequency]` was also provided as input, we could also add the number of times each ASV was observed, and the number of samples each ASV was observed in.~~ 

Update: rather than taking a `FeatureTable` directly, this action should take feature metadata as input. This allows for arbitrary metadata to be integrated in the resulting visualization, including the number of times each ASV was observed, and the number of samples each ASV was observed in which will be new `ImmutableMetadata` outputs from `summarize` following #282. This information is useful for example to help users to prioritize ASVs that are common for further investigation.

Oddant1:
@gregcaporaso this should probably be closed by #284, but I'll let you look at things and close this or request further changes when you get the chance.

gregcaporaso:
@Oddant1, @Clockwork-Rat - thanks so much for the work on this. I downloaded and tested #284, and it's working great. 

A few (little?) requests before we close this one out:
1. I think we may have talked about this in person and I forgot to write it down in the issue. Could you drop the Confidence scores from this visualization? I don't think they're useful in this context. 
2. What do you think of making the `FeatureData[Taxonomy]` into a `List` type, so more than one could be added? This is often of interest when a user is comparing a few taxonomy databases or classifiers (e.g., I showed the .qzv I generated to @cherman2, and her first question was if she could provide more than one taxonomy). The hard part there is how to differentiate them in the visualization - maybe a starting point could be just to name them based on their order in the list if more than one is provided (e.g., `Taxon: 0`, `Taxon: 1`, ...). If only one FeatureData[Taxonomy] is provided as input, just leave it as `Taxon`. This isn't ideal, but I think the folks who would actually use this would be happy enough to have the functionality that they'd be willing to track this on their own until we have a better solution.
[...]

PR #285, opened 2023-09-18
Summarize #276

Clockwork-Rat:
Update to reflect new additions in #276


q2_feature_table/tests/test_summarize.py
```diff
@@ -160,6 +160,200 @@ def test_tsv_builder(self):
...
+
+        try:
+            with tempfile.TemporaryDirectory() as output_dir:
+                tabulate_seqs(output_dir, seqs, metadata=metadata,
+                              taxonomy=taxonomy, merge_method="intersect")
+                # Did not error out, this is a problem
+                self.assertTrue(False)
+        except Exception:
+            # Should error out
+            self.assertTrue(True)
```
Oddant1: Use [this](https://docs.python.org/3.8/library/unittest.html#unittest.TestCase.assertRaisesRegex) for the exception

Clockwork-Rat:
@Oddant1 Everything should be good here, it's been tested

q2_feature_table/tests/test_summarize.py
```diff
@@ -210,10 +211,148 @@ def test_optional_inputs(self):
...
+                                       'seq05', 'seq90',
+                                       'seq48', 'seq08'],
+                                columns=['att1', 'att2'],
+                                data=[['00', '01'], ['10', '11'],
+                                      ['03', '04'], ['12', '13'],
+                                      ['05', '06'], ['14', '15'],
+                                      ['07', '08'], ['16', '17']])
+        metadata.index.name = 'feature id'
+
+        taxonomy = pd.DataFrame([('a;b;c;d', '1.0'), ('a;b;c;f', '0.7'),
```
Oddant1, line 241: Goes for all tests. Since we are accepting Collections of taxonomies now, probably split this into two taxonomies and put them in a dict then pass that dict into the test.

q2_feature_table/tests/test_summarize.py
```diff
@@ -210,10 +211,148 @@ def test_optional_inputs(self):
...
+                                 ('a;b;h;d', '0.3'), ('a;b;d;f', '0.7'),
+                                 ('a;b;e;d', '0.4'), ('a;b;c;f', '0.6'),
+                                 ('a;b;t;d', '1.0'), ('a;b;d;f', '0.5')],
+                                index=['seq17', 'seq02', 'seq03', 'seq48',
+                                       'seq05', 'seq19', 'seq07', 'seq08'],
+                                columns=['Taxon', 'Confidence'])
+
+        metadata = qiime2.Metadata(metadata)
+        taxonomy = pd.Series([taxonomy])
```
Oddant1: Also goes for all tests. Don't cast the taxonomies to a series keep them as dataframes.

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -31,21 +31,47 @@
 
 
 def tabulate_seqs(output_dir: str, data: DNAIterator,
-                  taxonomy: pd.DataFrame = None,
-                  metadata: qiime2.Metadata = None) -> None:
-    sequences = []
+                  taxonomy: pd.Series = None,
```
Oddant1: This is still going to need to be pd.Dataframe

q2_feature_table/_summarize/tabulate_seqs_assets/index.html
```diff
@@ -92,8 +92,8 @@ <h3>Sequence Table</h3>
             <th data-tsorter="numeric">Sequence Length</th>
             <th data-tsorter="link">Sequence</th>
             {% if taxonomy is defined %}
-              {% for col in taxonomy.columns %}
-            <th>{{ col }}</th>
+              {% for i in range(taxonomy.size) %}
+            <th>Taxon {{i}}</th>
```
Oddant1: Now just going to loop over `taxonomy.keys()`

q2_feature_table/plugin_setup.py
```diff
@@ -509,14 +509,17 @@
 plugin.visualizers.register_function(
     function=q2_feature_table.tabulate_seqs,
     inputs={'data': FeatureData[Sequence | AlignedSequence],
-            'taxonomy': FeatureData[Taxonomy]},
-    parameters={'metadata': Metadata},
+            'taxonomy': List[FeatureData[Taxonomy]]},
```
Oddant1: Actually make this `Collection[FeatureData[Taxonomy]]` that way they can key the taxonomies if they want.

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -31,21 +31,47 @@
...
+        for member in taxonomy:
+            if merge_method == 'union':
+                display_sequences = display_sequences.union(member.index)
+            elif merge_method == 'intersect':
+                display_sequences = display_sequences.intersection(
+                    member.index)
+            elif merge_method == 'strict':
+                if set(member.index) != display_sequences:
+                    raise Exception('Merge method is strict and IDs do not \
+                        match')
```
Oddant1: Taxonomy will now be a dict of dataframes at this point. Currently the loop is looping through a series so each member is an entry in the series. You're going to want to loop over the values in the dictionary `taxonomy.values()`

q2_feature_table/tests/test_summarize.py
```diff
@@ -210,10 +211,148 @@ def test_optional_inputs(self):
...
+                self.assertTrue('<td>1</td>' in file_text)
+                self.assertTrue('<td>1</td>' in file_text)
+                self.assertTrue('<td>2</td>' in file_text)
+                self.assertTrue('<td>3</td>' in file_text)
+                self.assertTrue('<td>4</td>' in file_text)
+                self.assertTrue('<td>6</td>' in file_text)
+                self.assertTrue('<td>a;b;e;d</td>' in file_text)
+                self.assertTrue('<td>a;b;c;d</td>' in file_text)
+                self.assertTrue('<td>08</td>' in file_text)
+                self.assertTrue('<td>17</td>' in file_text)
```
Oddant1: Again, goes for all tests, probably assert that the string "Taxonomy {key for the taxonomy}" exists in the output now.

q2_feature_table/_summarize/tabulate_seqs_assets/index.html
```diff
@@ -104,20 +104,35 @@ <h3>Sequence Table</h3>
...
+            <td></td>
+            {% endif %}
             {% if taxonomy is defined %}
-              {% for col in taxonomy.columns %}
-            <td>{{ taxonomy.loc[sequence.id, col] }}</td>
+              {% for member in taxonomy %}
+                {% if sequence in member.index %}            
+            <td>{{ member.loc[sequence, "Taxon"] }}</td>
+                {% else %}
+            <td></td>
```
Oddant1: Going to be looping over `taxonomy.values()`

Oddant1 (changes requested):

Oddant1:
If only I could order the comments lol. Basically let's make it a `Collection[FeatureData[Taxonomy]]` not a list that way they can name their entries if they want. Most other comments fall out from that,

gregcaporaso:
@Oddant1, @Clockwork-Rat - I have a note on my to-do list to test this out, but it looks like there are some comments still to be addressed. Can you ping me when those are addressed? I'll do my user-level testing/review then.

q2_feature_table/tests/test_summarize.py
```diff
@@ -210,10 +212,149 @@ def test_optional_inputs(self):
...
+                                 ('a;b;h;d', '0.3'), ('a;b;d;f', '0.7'),
+                                 ('a;b;e;d', '0.4'), ('a;b;c;f', '0.6'),
+                                 ('a;b;t;d', '1.0'), ('a;b;d;f', '0.5')],
+                                index=['seq17', 'seq02', 'seq03', 'seq48',
+                                       'seq05', 'seq19', 'seq07', 'seq08'],
+                                columns=['Taxon', 'Confidence'])
+
+        metadata = qiime2.Metadata(metadata)
+        taxonomy = {"Taxon Name": [taxonomy]}
+        taxonomy = pd.DataFrame.from_dict(taxonomy)
```
Oddant1: Test data for taxonomy should be a dictionary of dataframes. Something like:

```
taxonomy = {
    'foo': pd.DataFrame([('a;b;c;d', '1.0'), ('a;b;c;f', '0.7'),
                                 ('a;b;h;d', '0.3'), ('a;b;d;f', '0.7')],
                                index=['seq17', 'seq02', 'seq03', 'seq48'],
                                columns=['Taxon', 'Confidence'])
    'bar': pd.DataFrame([('a;b;e;d', '0.4'), ('a;b;c;f', '0.6'),
                                 ('a;b;t;d', '1.0'), ('a;b;d;f', '0.5')],
                                index=['seq17', 'seq02', 'seq03', 'seq48''],
                                columns=['Taxon', 'Confidence'])
}
```

Then just pass in that dictionary and assert that `Taxon foo` and `Taxon bar` are in the final viz. 

For union I didn't put all the IDs in both of them but for strict all the IDs would need to be in each of them obviously. The idea here is that we have n different taxonomies that came from different classifiers and we want to put all of them in the viz together.

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -32,20 +32,47 @@
...
+            if merge_method == 'union':
+                display_sequences = display_sequences.union(
+                    taxonomy[name][0].index)
+            elif merge_method == 'intersect':
+                display_sequences = display_sequences.intersection(
+                    taxonomy[name][0].index)
+            elif merge_method == 'strict':
+                if set(taxonomy[name][0].index) != display_sequences:
+                    raise Exception('Merge method is strict and IDs do not \
+                        match')
```
Oddant1: Something like

```
for value in taxonomy.values()
```

then value is a dataframe so you can just do

```
value.index
```

everywhere you get the index

Oddant1 (changes requested):

q2_feature_table/_summarize/_visualizer.py
Clockwork-Rat: That doesn't work, taxonomy.values returns a numpy representation of the same data, I tried that already, but the dataframe will always have one row of values at index zero, unless the user somehow made a mistake.

q2_feature_table/tests/test_summarize.py
Clockwork-Rat: This is effectively what I have done, but I can add more taxonomy data if you would like. Right now there is only one taxonomy, but you would like me to add more? I can definitely do that.

gregcaporaso:
@Clockwork-Rat, @Oddant1 - I'm doing some testing of `tabulate-seqs` and it's working great. I'm experimenting with variations of the following command with the attached files ([tab-seqs-tests.zip](https://github.com/qiime2/q2-feature-table/files/12703447/tab-seqs-tests.zip)). 

```bash
qiime feature-table tabulate-seqs \
 --i-data rep-seqs.qza \
 --i-taxonomy gg:taxonomy.qza \
 --o-visualization rep-seqs.qzv \
 --i-taxonomy something_else:taxonomy.qza \
--m-metadata-file feature-frequency-detail.csv
```

A few quick requests:
- add usage example that provides a single taxonomy - let's not name the collection input in this case (e.g., `--i-taxonomy taxonomy.qza`)
- add usage example that provides two taxonomies, and illustrate how to give them descriptive names in the example (e.g., `--i-taxonomy taxonomy_a:taxonomy-1.qza --i-taxonomy taxonomy_b:taxonomy-2.qza`). for the purpose of this example, those two inputs could be silva and greengenes taxonomies, or greengenes 13_8 and greengenes2).
- validate `merge-method` value (though that might be a different PR)? - I accidentally ran the following command, and was confused about why I wasn't see the metadata in the output. it's b/c i accidentally provided `--p-merge-method` instead of `--m-metadata-file`
```
# commands succeeds when receiving --p-merge-method but not --m-metadata-file
qiime feature-table tabulate-seqs --i-data rep-seqs.qza --i-taxonomy gg:taxonomy.qza --p-merge-method feature-frequency-detail.csv --o-visualization rep-seqs.qzv --i-taxonomy something_else:taxonomy.qza
Saved Visualization to: rep-seqs.qzv

# Then, I added --m-metadata-file, and it still succeeded - should definitely have thrown an error here:
qiime feature-table tabulate-seqs --i-data rep-seqs.qza --i-taxonomy gg:taxonomy.qza --p-merge-method feature-frequency-detail.csv --o-visualization rep-seqs.qzv --i-taxonomy something_else:taxonomy.qza --m-metadata-file feature-frequency-detail.csv
```

I'll poke around a bit more and follow up if I notice anything else (just need to run for a meeting). (**EDIT: Done - no additional feedback!**)

Great work - this is going to be _so handy_ to have!

gregcaporaso:
Weirdly, I just noticed the column headers for the `Sequence` and `Sequence length` columns are mixed up. Not sure if that was there before or introduced here, but we should fix that in this PR. 

Screenshot 2023-09-22 at 11 00 19 AM

Oddant1:
@gregcaporaso that should all be doable including the validation on `--p-merge-method`. I probably should have noticed that. We can just add `% Choices`

Oddant1 (approved):

PR #286, opened 2023-09-19
Fixes #149

### LINKED ISSUE #149: BIOM 2.1.6 refactors
**Improvement Description**
On a review of the codebase, it looks like the following can be refactored to take advantage of new endpoints available in BIOM 2.1.6 once #148 is merged.

- [x] `_filter._filter_table` can use `Table.remove_empty() on these [lines](https://github.com/qiime2/q2-feature-table/blob/2ac9768c94484b82d2397953c16b4560a4b25647/q2_feature_table/_filter.py#L56-L61).
- [x] `_merge.merge` can use `Table.concat`. For two tables, this will be faster. `concat` supports an arbitrary number of `Table` instances though, so API call will become important once variadic arguments are supported (e.g., to allow the user to merge N tables).
- [x] `_sample.sample can use `Table.remove_empty()` [here](https://github.com/qiime2/q2-feature-table/pull/146/files#diff-bf9aec931e0643ae3dfee4bb85f5bb1bR22). Note as of the creation of this issue, this method was in an unmerged PR #146 
 
**References**
1. #148 
2. [lines](https://github.com/qiime2/q2-feature-table/blob/2ac9768c94484b82d2397953c16b4560a4b25647/q2_feature_table/_filter.py#L56-L61)
3. [here](https://github.com/qiime2/q2-feature-table/pull/146/files#diff-bf9aec931e0643ae3dfee4bb85f5bb1bR22)
4. #146

wasade:
Fixes #149. Note merge items reflected in #149 were resolved in a different PR it seems. Added use of `partition` for `split`.

### DIFF
--- a/q2_feature_table/_split.py
+++ b/q2_feature_table/_split.py
@@ -14,12 +14,14 @@ def split(table: biom.Table,
           metadata: qiime2.CategoricalMetadataColumn,
           filter_empty_features: bool = True) -> biom.Table:
     metadata_df = metadata.drop_missing_values().to_dataframe()
+    lookup = metadata_df[metadata.name].to_dict()
 
-    indices = metadata_df.reset_index(
-        ).groupby(metadata.name)[metadata_df.index.name].apply(list).to_dict()
+    def partition_f(i, m):
+        return lookup.get(i)
 
+    unique_grps = sorted(set(lookup.values()))
     try:
-        qiime2.sdk.util.validate_result_collection_keys(*indices.keys())
+        qiime2.sdk.util.validate_result_collection_keys(*unique_grps)
     except KeyError as e:
         raise KeyError(
             "One or more invalid metadata column values identified during "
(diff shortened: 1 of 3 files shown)


gregcaporaso:
Thanks @wasade! Does biom.partition now handle the missing metadata case we were discussing [here](https://github.com/qiime2/q2-feature-table/pull/277#discussion_r1276760250)? I assume so since the unit tests are passing, and I do have a test for that case. (Sorry for not investigating myself, I'm really tied up this morning and I figured you know offhand.)

q2_feature_table/_split.py
```diff
@@ -28,9 +30,11 @@ def split(table: biom.Table,
             f"table. The original error message is as follows: {str(e)}")
 
     result = {}
-    for group, sample_ids in indices.items():
-        t = table.filter(sample_ids, axis='sample', inplace=False)
+    for group, tab in table.partition(partition_f):
+        if group is None:
+            continue
```
wasade, line 35: @gregcaporaso, we simply ignore tables where a null group was observed. The null is exposed via the `.get` in the `partition_f`. Does that seem reasonable for the intended uses?

gregcaporaso: Just to make sure I understand: you generate the partitioned table containing all samples that are missing metadata (i.e., the value is null), but just don't do anything with it so it's not included in the output? If so, that seems fine to me.

wasade: Correct. Specifically, if the sample ID is not in the metadata lookup, a `None` is returned. The `None` group is then ignored.

gregcaporaso:
Replacing this with #323, which addresses the merge conflicts. That one is ready for merge when the tests pass. Sorry this one took forever to get to @wasade. Note that we'll integrate the new `Table.partion` API features when we update to biom 2.1.16 (planned for this release cycle; #324).

PR #288, opened 2023-10-04
Summarize pipeline 282

### LINKED ISSUE #282: add new outputs to `summarize`
The information in the `Interactive Sample Detail` and `Feature Detail` tabs of the visualization are useful for integration in downstream analyses. This information can currently be downloaded from the `Overview` tab of the `.qzv` (where it's almost, but not quite in our metadata format). 

Now that we have the `ImmutableMetadata` semantic type, we should output this information in two new outputs. The first should contain the information from the `Interactive Sample Detail` tab, which will be sample metadata, and the second should contain the information from the `Feature Detail` tab. This will allow for integration of these data anywhere that metadata is accepted in QIIME 2 (e.g., this makes it easy to test for a correlation between alpha diversity and samples' total frequencies, or to view features' sequences, taxonomy, total frequency, and the number of samples they're observed in in a single visualization).

This will necessitate transitioning `summarize` from a `Visualizer` to a `Pipeline`. 

**EDIT: It also probably makes the most sense to create a new `Action` (or two) that generates the information described above from a `FeatureTable` in different `ImmutableMetadata` outputs. Then the `Pipeline` would call that `Action(s)`. The feature-focused `Action` could be called `tabulate-feature-frequencies` (which outputs one `ImmutableMetadata` with columns for the total count of each feature (i.e., sum of each column, indexed on the column headers) and the total number of samples each feature is observed in (i.e., count of non-zero values in each column, indexed on the column headers). The sample-focused `Action` could be called `tabulate-sample-frequencies`, which outputs the same information but focused on the samples instead of the features. The count of non-zero values would be the same as computing *observed features* on the table, but I think it could be helpful to have this information before rarefaction/normalization of the table (e.g., to see if read counts obtained per sample drive separation in PCoA, even after rarefaction/normalization).**

Open question: adding two new required outputs will be a breaking change to the interface here. Is that an acceptable breaking change, or should we make these optional outputs for one release cycle? 

This feature request has come up before in the context of these issues:
https://github.com/qiime2/q2-feature-table/issues/161
https://github.com/qiime2/q2-feature-table/issues/158

Clockwork-Rat:
Closes #282 
Add tabulate frequencies actions to q2-feature-table and create pipeline that calls tabulate frequencies and summarize called summarize plus


q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +653,37 @@
...
+    outputs={'metadata': ImmutableMetadata},
+    input_descriptions={
+        'table': 'A table containing feature frequency information'
+    },
+    output_descriptions={
+        'metadata': 'Metadata generated from frequency data'
+    },
+    name='Tabulate feature frequencies',
+    description='Tabulates feature frequencies from a feature table'
+                'and generates metadata',
```
Oddant1: Need a space either at the beginning of this string or end of prior string.

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +653,37 @@
                 'splits are defined by values in metadata column.',
     examples={}
 )
+
+plugin.actions.register_function(
```
Oddant1: plugin.methods not .actions

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +653,37 @@
...
+    output_descriptions={
+        'metadata': 'Metadata generated from frequency data'
+    },
+    name='Tabulate feature frequencies',
+    description='Tabulates feature frequencies from a feature table'
+                'and generates metadata',
+    examples={}
+)
+
+plugin.actions.register_function(
```
Oddant1: Same as other registration

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +653,37 @@
...
+    outputs={'metadata': ImmutableMetadata},
+    input_descriptions={
+        'table': 'A table containing sample frequency information'
+    },
+    output_descriptions={
+        'metadata': 'Metadata generated from frequency data'
+    },
+    name='Tabulate sample frequencies',
+    description='Tabulates feature frequencies from a feature table'
+                'and generates metadata',
```
Oddant1: Same as other description

Oddant1 (changes requested):

q2_feature_table/tests/test_summarize.py
```diff
@@ -516,5 +518,23 @@ def test_vega_spec_nandling(self):
...
+
+class TabulateFeatureFrequencyTests(TestCase):
+
+    def test_basic_case(self):
+        table = biom.Table(np.array([[25, 25, 25], [25, 25, 25]]),
+                           ['O1', 'O2'],
+                           ['S1', 'S2', 'S3'])
+        tabulate_feature_frequencies(table)
+
+
```
Oddant1: Need to add actual assertions to the test. Just assert that the result of the action contains the data it should. Let me know if you need help with that

Oddant1:
Ok tests look good now

Oddant1:
Closes #282

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -217,6 +217,38 @@ def summarize(output_dir: str, table: biom.Table,
...
+    sample_frequencies = _frequencies(table, 'sample')
+    sample_frequencies = sample_frequencies.apply(
+        '{:,}'.format).to_frame('Frequency')
+    sample_frequencies.index.name = "Sample ID"
+    return qiime2.Metadata(sample_frequencies)
+
+
+def summarize_plus(ctx, output_dir: str, table: biom.Table,
+                   metadata: qiime2.Metadata = None):
+    _feature_frequencies = ctx.get_action('feature_table',
```
Oddant1: Get rid of the output_dir argument here and remove the type annotations

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,59 @@
...
+             'sample_metadata': ImmutableMetadata,
+             'visualized_data': Visualization},
+    input_descriptions={
+        'table': 'The feature table to be summarized.'
+    },
+    parameter_descriptions={'sample_metadata': 'The sample metadata.'},
+    output_descriptions={'frequency_metadata': 'Metadata pertaining to '
+                         'feature frequencies',
+                         'sample_metadata': 'Metadata pertaining to '
+                         'sample frequencies',
```
Oddant1: Standardize the language here with the language used in the individual actions. Right now this says "pertaining to" and the actions say "generated from." I'd probably go with "generated from" for both of them.

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,59 @@
...
+    description='Tabulates feature frequencies from a feature table '
+                'and generates metadata',
+    examples={}
+)
+
+plugin.methods.register_function(
+    function=q2_feature_table.tabulate_sample_frequencies,
+    inputs={'table': FeatureTable[Frequency]},
+    parameters={},
+    outputs={'metadata': ImmutableMetadata},
```
gregcaporaso: Let's call this `sample_frequencies`, here and throughout. `metadata` is too overloaded for it to be meaningful to our users (they won't really think of this as metadata).

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,59 @@
...
+plugin.methods.register_function(
+    function=q2_feature_table.tabulate_feature_frequencies,
+    inputs={'table': FeatureTable[Frequency]},
+    parameters={},
+    outputs={'metadata': ImmutableMetadata},
+    input_descriptions={
+        'table': 'A table containing feature frequency information'
+    },
+    output_descriptions={
+        'metadata': 'Metadata generated from frequency data'
```
gregcaporaso: Let's call this `feature_frequencies` here, and throughout (unless you guys object to this choice of names). We should avoid referring to these as "metadata" in user-facing documentation, as this won't be something that our users initially associate as metadata so it will cause confusion. (Same goes for the "sample_metadata")

gregcaporaso:
```suggestion
        'feature_frequencies': 'Per-sample and total frequencies per feature.'
```

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,59 @@
...
     examples={}
 )
+
+plugin.methods.register_function(
+    function=q2_feature_table.tabulate_feature_frequencies,
+    inputs={'table': FeatureTable[Frequency]},
+    parameters={},
+    outputs={'metadata': ImmutableMetadata},
+    input_descriptions={
+        'table': 'A table containing feature frequency information'
```
gregcaporaso:
```suggestion
        'table': 'The input feature table.'
```

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,59 @@
...
+    },
+    output_descriptions={
+        'metadata': 'Metadata generated from frequency data'
+    },
+    name='Tabulate feature frequencies',
+    description='Tabulates feature frequencies from a feature table '
```
gregcaporaso: description: 'Tabulate sample count and total frequency per feature.'

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -217,6 +217,37 @@ def summarize(output_dir: str, table: biom.Table,
     plt.close('all')
 
 
+def tabulate_feature_frequencies(table: biom.Table) -> qiime2.Metadata:
+    feature_frequencies = _frequencies(table, 'observation')
+    feature_frequencies = feature_frequencies.apply(
+        '{:,}'.format).to_frame('Frequency')
```
gregcaporaso, line 217: The Metadata output from this function should also contain the total number of samples each feature is observed in (so one more column in the output).

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -217,6 +217,37 @@ def summarize(output_dir: str, table: biom.Table,
...
+    feature_frequencies.index.name = "Feature ID"
+    return qiime2.Metadata(feature_frequencies)
+
+
+def tabulate_sample_frequencies(table: biom.Table) -> qiime2.Metadata:
+    sample_frequencies = _frequencies(table, 'sample')
+    sample_frequencies = sample_frequencies.apply(
```
gregcaporaso, line 228: The Metadata output from this function should additionally contain the total number of features observed in each sample (one more column in the output).

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -217,6 +217,37 @@ def summarize(output_dir: str, table: biom.Table,
     plt.close('all')
 
 
+def tabulate_feature_frequencies(table: biom.Table) -> qiime2.Metadata:
```
gregcaporaso, line 214: I'm thinking this should be called `feature_counts`, since it's a few different values per feature, not just frequencies (see my next comment). @Oddant1, does that name make sense to you?

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -217,6 +217,37 @@ def summarize(output_dir: str, table: biom.Table,
...
+    feature_frequencies = _frequencies(table, 'observation')
+    feature_frequencies = feature_frequencies.apply(
+        '{:,}'.format).to_frame('Frequency')
+    feature_frequencies.index.name = "Feature ID"
+    return qiime2.Metadata(feature_frequencies)
+
+
+def tabulate_sample_frequencies(table: biom.Table) -> qiime2.Metadata:
```
gregcaporaso, line 226: I'm thinking this should be called `sample_counts`, since it's a few different values per sample, not just frequencies. @Oddant1, does that name make sense to you?

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -217,6 +217,37 @@ def summarize(output_dir: str, table: biom.Table,
...
+
+def summarize_plus(ctx, table, metadata=None):
+    _feature_frequencies = ctx.get_action('feature_table',
+                                          'tabulate_feature_frequencies')
+    _sample_frequencies = ctx.get_action('feature_table',
+                                         'tabulate_sample_frequencies')
+    _visualizer = ctx.get_action('feature_table',
+                                 'summarize')
+
+    feature_metadata, = _feature_frequencies(table)
```
gregcaporaso: feature_frequencies = ...

q2_feature_table/tests/test_summarize.py
```diff
@@ -516,5 +518,33 @@ def test_vega_spec_nandling(self):
...
 
+class TabulateSampleFrequencyTests(TestCase):
+
+    def test_basic_case(self):
+        table = biom.Table(np.array([[25, 25, 25], [25, 25, 25]]),
+                           ['O1', 'O2'],
+                           ['S1', 'S2', 'S3'])
+        metadata = tabulate_sample_frequencies(table).to_dataframe()
+
+        self.assertTrue('Frequency' in metadata.columns)
```
gregcaporaso: You can generate the exact DataFrame that you expect as output here, since the input is so simple. Can you do that instead? 

Also, be sure to test for edge cases (e.g, empty feature table either creates empty output, or raises an error - whichever you guys think makes more sense).

q2_feature_table/tests/test_summarize.py
```diff
@@ -18,7 +18,9 @@
 from q2_types.feature_data import DNAIterator
 import csv
 
-from q2_feature_table import tabulate_seqs, summarize
+from q2_feature_table import (
```
gregcaporaso, line 23: We should have tests of `summarize_plus` as well.

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,59 @@
                 'splits are defined by values in metadata column.',
     examples={}
 )
+
+plugin.methods.register_function(
+    function=q2_feature_table.tabulate_feature_frequencies,
+    inputs={'table': FeatureTable[Frequency]},
```
gregcaporaso: Let's have all these actions work on `FeatureTable[PresenceAbsence]` as well, as these are going to become more common in some of our workflows.

gregcaporaso (changes requested):
@Oddant1, @Clockwork-Rat - let me know if you have questions. I'm suggesting some different nomenclature, and I may not have hit ever place where the original nomenclature you're using shows up - just be sure to take a full pass through to make sure that the nomenclature is consistent.

q2_feature_table/tests/test_summarize.py
Oddant1: Ok, yeah this was one thing I was wondering about. Are there any edge cases here other than "empty?"

q2_feature_table/tests/test_summarize.py
Oddant1: It almost feels redundant testing this since it just calls 3 actions and doesn't use the output from one as input to another it just calls them straight. Do you think we should bother covering all cases for `summarize_plus` or just the base case?

q2_feature_table/_summarize/_visualizer.py
Oddant1: Same as above, sounds good to me

q2_feature_table/_summarize/_visualizer.py
Oddant1: Sounds good to me

q2_feature_table/tests/test_summarize.py
```diff
@@ -516,5 +518,33 @@ def test_vega_spec_nandling(self):
...
+
+    def test_basic_case(self):
+        table = biom.Table(np.array([[25, 25, 25], [25, 25, 25]]),
+                           ['O1', 'O2'],
+                           ['S1', 'S2', 'S3'])
+        metadata = tabulate_feature_frequencies(table).to_dataframe()
+        self.assertTrue('Frequency' in metadata.columns)
+        self.assertTrue('O1' in metadata.index)
+        self.assertTrue('O2' in metadata.index)
+        self.assertTrue('75.0' in str(metadata))
```
Oddant1: Here same as above from @gregcaporaso. Create the dataframe you expect to see. Something like

```
exp = pd.Dataframe("whatever data")
obs = # The dataframe you actually received as a result
pd.teseting.assert_frames_equal(exp, obs)
```

Documentation on assert_frames_equal [here](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html)

q2_feature_table/plugin_setup.py
gregcaporaso: Update: Let's have this work on `FeatureTable[RelativeFrequency]` too, so it matches all the input types of `summarize`.

q2_feature_table/tests/test_summarize.py
gregcaporaso: Just a simple base case will be fine.

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,59 @@
...
+        'metadata': 'Metadata generated from frequency data'
+    },
+    name='Tabulate sample frequencies',
+    description='Tabulates feature frequencies from a feature table '
+                'and generates metadata',
+    examples={}
+)
+
+plugin.pipelines.register_function(
+    function=q2_feature_table.summarize_plus,
```
gregcaporaso, line 696: As it stands, I think this will fail (or should at least) on the `FeatureTable` types other than `Frequency`, since the `Pipeline` allows all three as input and the two new actions only allow `Frequency` as input.

q2_feature_table/_summarize/_visualizer.py
gregcaporaso: I changed my mind on this, after thinking about it a little more. Let's stick with `tabulate_feature_frequencies` and `tabulate_sample_frequencies` for the action names.

q2_feature_table/_summarize/_visualizer.py
gregcaporaso: Changed my mind, let's stick with `tabulate_sample_frequencies`.

q2_feature_table/tests/test_summarize.py
gregcaporaso: Not that is coming to mind right now, though at least the underlying actions should be tested with the three input table types.

q2_feature_table/tests/test_summarize.py
Clockwork-Rat: Sounds good, I'll create this right now

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,61 @@
...
+    outputs={'sample_frequencies': ImmutableMetadata},
+    input_descriptions={
+        'table': 'The input feature table.'
+    },
+    output_descriptions={
+        'sample_frequencies': 'Per-feature and total frequencies per sample.'
+    },
+    name='Tabulate sample frequencies',
+    description='Tabulate sample count and total frequency per feature.'
+                'and generates metadata',
```
Oddant1:
```suggestion
    description='Tabulate feature count and total frequency per sample.'
```

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,61 @@
...
+    outputs={'feature_frequencies': ImmutableMetadata},
+    input_descriptions={
+        'table': 'The input feature table.'
+    },
+    output_descriptions={
+        'feature_frequencies': 'Per-sample and total frequencies per feature.'
+    },
+    name='Tabulate feature frequencies',
+    description='Tabulates feature frequencies from a feature table '
+                'and generates metadata',
```
Oddant1:
```suggestion
    description='Tabulate sample count and total frequency per feature.'
```

q2_feature_table/_examples.py
```diff
@@ -388,3 +388,49 @@ def feature_table_tabulate_seqs_multi_taxon(use):
...
+        use.UsageAction('feature_table', 'summarize_plus'),
+        use.UsageInputs(table=feature_table),
+        use.UsageOutputNames(feature_frequencies='feature_frequencies',
+                             sample_frequencies='sample_frequencies',
+                             visualized_data='visual summary')
+    )
+
+    feature_freqs.assert_output_type('ImmutableMetadata')
+    sample_freqs.assert_output_type('ImmutableMetadata')
+    viz.assert_output_type('ImmutableMetadata')
```
Oddant1: The viz is a `Visualization` not `ImmutableMetadata`

q2_feature_table/_examples.py
```diff
@@ -388,3 +388,49 @@ def feature_table_tabulate_seqs_multi_taxon(use):
...
+
+    viz.assert_output_type('ImmutableMetadata')
+
+
+def feature_table_tabulate_feature_freqs(use):
+    feature_table = use.init_artifact_from_url(
+        'feature_table', moving_pics_ft_url
+    )
+
+    viz, = use.action(
```
Oddant1: This is `feature_freqs` not `viz`

q2_feature_table/_examples.py
```diff
@@ -388,3 +388,49 @@ def feature_table_tabulate_seqs_multi_taxon(use):
...
 
     viz.assert_output_type('Visualization')
+
+
+def feature_table_tabulate_sample_freqs(use):
+    feature_table = use.init_artifact_from_url(
+        'feature_table', moving_pics_ft_url
+    )
+
+    viz, = use.action(
```
Oddant1: This is `sample_freqs` not `viz`

Oddant1 (changes requested):

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,62 @@
...
+                                  PresenceAbsence]},
+    parameters={'metadata': Metadata},
+    outputs={'feature_frequencies': ImmutableMetadata,
+             'sample_frequencies': ImmutableMetadata,
+             'visualized_data': Visualization},
+    input_descriptions={
+        'table': 'The feature table to be summarized.'
+    },
+    parameter_descriptions={'metadata': 'The sample metadata.'},
+    output_descriptions={'feature_frequencies': 'Metadata generated from ' +
```
gregcaporaso: Drop all descriptions of this as `Metadata`, as that's not going to be intuitive to some users. 

Read this description as if you didn't already know what this was - would it be helpful? Try to use that lens when reviewing all of your user-facing text. Please take a pass through all user-facing text.

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,62 @@
...
+                         'input table',
+                         'sample_frequencies': 'Metadata generated from ' +
+                         'input table',
+                         'visualized_data': 'Visual summary of feature table'},
+    name="Summarize table plus",
+    description="Generate visual and tabular summaries of a feature table. "
+                "Tabulate sample and feature frequencies as metadata",
```
gregcaporaso: You're missing a period at the end of the second sentence here (and also, see my comment above about "metadata").

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,62 @@
...
+             'visualized_data': Visualization},
+    input_descriptions={
+        'table': 'The feature table to be summarized.'
+    },
+    parameter_descriptions={'metadata': 'The sample metadata.'},
+    output_descriptions={'feature_frequencies': 'Metadata generated from ' +
+                         'input table',
+                         'sample_frequencies': 'Metadata generated from ' +
+                         'input table',
+                         'visualized_data': 'Visual summary of feature table'},
```
gregcaporaso: Let's call this output `summary`.

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,62 @@
...
+    function=q2_feature_table.tabulate_sample_frequencies,
+    inputs={'table': FeatureTable[Frequency | PresenceAbsence |
+                                  RelativeFrequency]},
+    parameters={},
+    outputs={'sample_frequencies': ImmutableMetadata},
+    input_descriptions={
+        'table': 'The input feature table.'
+    },
+    output_descriptions={
+        'sample_frequencies': 'Per-feature and total frequencies per sample.'
```
gregcaporaso:
```suggestion
        'sample_frequencies': 'Observed feature count and total frequencies per sample.'
```

You can use this as the description for output from `summarize-plus` as well.

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,62 @@
...
+    },
+    name='Tabulate sample frequencies',
+    description='Tabulate sample count and total frequency per feature.',
```
gregcaporaso: This is the description for `tabulate_feature_frequencies`

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,62 @@
...
+    function=q2_feature_table.tabulate_feature_frequencies,
+    inputs={'table': FeatureTable[Frequency | PresenceAbsence |
+                                  RelativeFrequency]},
+    parameters={},
+    outputs={'feature_frequencies': ImmutableMetadata},
+    input_descriptions={
+        'table': 'The input feature table.'
+    },
+    output_descriptions={
+        'feature_frequencies': 'Per-sample and total frequencies per feature.'
```
gregcaporaso, line 668: You can use this as the description for output from `summarize-plus` as well.

q2_feature_table/plugin_setup.py
```diff
@@ -652,3 +654,62 @@
...
+    },
+    name='Tabulate feature frequencies',
+    description='Tabulate feature count and total frequency per sample.',
```
gregcaporaso: This is the description for `tabulate_sample_frequencies`

q2_feature_table/tests/test_summarize.py
```diff
@@ -516,5 +490,77 @@ def test_vega_spec_nandling(self):
         self.assertEqual(spec['data'][0]['values'], exp)
 
 
+class TabulateSampleFrequencyTests(TestCase):
+
+    def test_basic_case(self):
+        table = biom.Table(np.array([[25, 25, 25], [25, 25, 25]]),
```
gregcaporaso: Can you make this table more variable? If the underlying code always returned 50 and 2, this test would pass. Make some samples have a different total frequency, and make some features present in only some of the samples.

q2_feature_table/tests/test_summarize.py
```diff
@@ -516,5 +490,77 @@ def test_vega_spec_nandling(self):
...
+                            ['2', '2', '2']},
+                           index=['S1', 'S2', 'S3'])
+        exp.index.name = 'Sample ID'
+        pd.testing.assert_frame_equal(exp, obs)
+
+
+class TabulateFeatureFrequencyTests(TestCase):
+
+    def test_basic_case(self):
+        table = biom.Table(np.array([[25, 25, 25], [25, 25, 25]]),
```
gregcaporaso: Same request as above - make the features not all identical, in terms of both pieces of information you're returning about them.

gregcaporaso: Also, does this handle empty tables? What about empty samples or empty features in those tables? Add tests for these edge cases to all tests.

q2_feature_table/tests/test_summarize.py
```diff
@@ -516,5 +490,77 @@ def test_vega_spec_nandling(self):
...
+class SummarizePlusTests(TestPluginBase):
+
+    package = 'q2_feature_table'
+
+    def setUp(self):
+        super().setUp()
+        self.summarize_plus = self.plugin.pipelines['summarize_plus']
+
+    def test_basic(self):
+        table = biom.Table(np.array([[25, 25, 25], [25, 25, 25]]),
```
gregcaporaso: Same comment as above for the `summarize-plus` tests.

gregcaporaso (changes requested):
@Clockwork-Rat, a few last requests. I tested locally and it all seems to be working great. 

@Oddant1, have you reviewed all of the code in the new actions? I'm assuming you had, so I focused on the user-facing aspects in this review. If you haven't reviewed all of the code in the new actions already, could you take a pass?

q2_feature_table/tests/test_summarize.py
Clockwork-Rat: @Oddant1 are there examples of what should happen in these edge cases?

Oddant1: @Clockwork-Rat it doesn't looke like we really cover these edge cases for summarize currently. [This](https://github.com/Clockwork-Rat/q2-feature-table/blob/174761b67f9522a17c5e79d3c2e5addd4f2a9e83/q2_feature_table/tests/test_summarize.py#L478) looks like the closest thing we currently have.

If you have empty samples or empty features you're probably expecting. NaNs or Nones in the relevant rows of the interactive sample detail and the feature detail tabs respectively. If the entire thing is empty you should probably just error out (I expect something would cause an error anyway as it stands but idk).

I would try putting your edge case data through the existing summarize viz and see what the results look like/what errors you get honestly. It doesn't look like there are any tests for that, so you can't tell exactly what you're expecting there by reading the tests.

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -236,6 +236,16 @@ def tabulate_sample_frequencies(table: biom.Table) -> qiime2.Metadata:
 
 
 def summarize_plus(ctx, table, metadata=None):
+
+    try:
+        table_dimensions = table.view(pd.DataFrame).shape
+
+    except ValueError:
+        raise ValueError('Cannot summarize a table with no Features')
```
gregcaporaso:
```suggestion
        raise ValueError('Cannot summarize a table with no features.')
```

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -236,6 +236,16 @@ def tabulate_sample_frequencies(table: biom.Table) -> qiime2.Metadata:
...
+
+    if table_dimensions[0] == 0:
+        raise ValueError('Cannot summarize a table with no Samples')
```
gregcaporaso:
```suggestion
        raise ValueError('Cannot summarize a table with no samples.')
```

gregcaporaso (approved):
This looks good to me. I added two suggestions on error messages - you should just be able to accept those suggestions. 

@Oddant1, anything else before this one is merged? Feel free to merge when you're ready. 

Thanks @Clockwork-Rat!

Oddant1:
lgtm

PR #294, opened 2023-11-27
Get blast URL correctly

Oddant1:
Fixes bug where hyperlinks on sequences were not linking to blast


cherman2:
The code changes look reasonable. This solution worked perfectly when I pulled this down locally and tested it. Thanks @Oddant1! 

I think this is ready to merge once the test pass!

PR #295, opened 2023-11-29
add formatting

### LINKED ISSUE #293: minor numeric formatting changes in `summarize` and `summarize-plus`
I noticed that a couple of long-standing formatting issues in `summarize` made their way into `summarize-plus`. @Clockwork-Rat, could you take care of these once #288 gets merged? 

First, in the tables on the first tab there are too many significant digits in the average - we should align those with the number of significant digits in the values in the table. 
Screenshot 2023-11-21 at 2 01 30 PM

Second, on the sample detail tab, we should format the frequencies with commas (as in the feature detail tab). 
Screenshot 2023-11-21 at 2 01 41 PM

Clockwork-Rat:
Closes #293

Adds formatting and rounds summary output.

### DIFF
--- a/q2_feature_table/_summarize/_visualizer.py
+++ b/q2_feature_table/_summarize/_visualizer.py
@@ -373,27 +373,14 @@ def is_integer(x):
 def _frequency_summary(table: biom.Table, axis='sample'):
     frequencies = _frequencies(table, axis=axis)
 
-    nonzeros = table.matrix_data.data
-
-    _round = True
-
-    for item in nonzeros:
-        if not is_integer(item):
-            _round = False
-            break
-
     first = frequencies.quantile(0.25)
     third = frequencies.quantile(0.75)
-    _fst = round(first, 1) if _round else round(first, 4)
-    _min = round(frequencies.min(), 1) if _round else round(
-        frequencies.min(), 4)
-    _thd = round(third, 1) if _round else round(third, 4)
-    _med = round(frequencies.median(), 1) if _round else round(
-        frequencies.median(), 4)
-    _max = round(frequencies.max(), 1) if _round else round(
-        frequencies.max(), 4)
-    mean = round(frequencies.mean(), 1) if _round else round(
-        frequencies.mean(), 4)
+    _fst = round(first, 1)
+    _min = round(frequencies.min(), 1)
+    _thd = round(third, 1)
+    _med = round(frequencies.median(), 1)
+    _max = round(frequencies.max(), 1)
+    mean = round(frequencies.mean(), 1)
 
     summary = pd.Series([_min, _fst,
                          _med, _thd,
--- a/q2_feature_table/_summarize/_visualizer.py
+++ b/q2_feature_table/_summarize/_visualizer.py
@@ -373,16 +373,13 @@ def is_integer(x):
 def _frequency_summary(table: biom.Table, axis='sample'):
     frequencies = _frequencies(table, axis=axis)
 
-    table = table.to_dataframe()
+    nonzeros = table.matrix_data.data
 
     _round = True
 
-    for _, row in table.iterrows():
-        for idx in row:
-            if not is_integer(idx):
-                _round = False
-                break
-        if not _round:
+    for item in nonzeros:
+        if not is_integer(item):
+            _round = False
             break
 
     first = frequencies.quantile(0.25)
--- a/q2_feature_table/_summarize/_visualizer.py
+++ b/q2_feature_table/_summarize/_visualizer.py
@@ -361,6 +361,15 @@ def _frequencies(table, axis):
     return pd.Series(data=table.sum(axis=axis), index=table.ids(axis=axis))
 
 
+def is_integer(x):
+    if isinstance(x, int):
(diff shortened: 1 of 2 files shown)


Clockwork-Rat:
@Oddant1 I still have to add number formatting to parts of the output, which I am working on now, but this is the basis and just adds the rounding. I am trying to see where I can do the rounding -- either in the jinja html or in the summary function. I am not sure which is better or more clear.

gregcaporaso:
I would advocate for doing the rounding in the Python function rather than the jinja html. My gut is that the former would be easier to test than the latter, and potentially more clear to future developers.

Clockwork-Rat:
@Oddant1 @gregcaporaso this should be ready to pull in.

gregcaporaso:
@Clockwork-Rat, should this be a Draft Pull Request until we sort out the plan for handling `FeatureTable[RelativeFrequency]`?

Clockwork-Rat:
@gregcaporaso @Oddant1 this should be working as expected.

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -356,12 +361,46 @@ def _frequencies(table, axis):
...
+        return False
+
+
+def _frequency_summary(table: biom.Table, axis='sample'):
     frequencies = _frequencies(table, axis=axis)
 
-    summary = pd.Series([frequencies.min(), frequencies.quantile(0.25),
-                         frequencies.median(), frequencies.quantile(0.75),
-                         frequencies.max(), frequencies.mean()],
+    table = table.to_dataframe()
```
ebolyen: Here's a better approach we can use:
```python
nonzeros = table.matrix_data.data
```
This will take the BIOM table to a Sparse Matrix, and then the raw non-zero data, which will prevent setting up a dataframe and iterating over rows or columns, since we don't care about where the values _are_, just their intergery-ness.

hagenjp:
@Clockwork-Rat, 
We are doing a PR sprint today and I was assigned to review this PR with the help of @ebolyen and @gregcaporaso. 
When rendering your changes locally using `feature-table summarize` we realized that relative frequency just does not make sense as an input of this method. In the attached screenshot the 2.13 maximum frequency stood out to us, summing across the relative frequency values is creating meaningless values inside the Frequency Per Feature table. 
In `summarize` and `summarize-plus` please drop relative frequency as an input option of these actions.  
Feel free to close this PR and open a new one, if you would like, and let me/us know if you need any clarification. (Evan and Greg apologize for not noticing this sooner, it was only after looking at the relative frequency summary that they realized the outcome here!)
Screen Shot 2023-12-11 at 11 01 06 AM

Clockwork-Rat:
@hagenjp @ebolyen These changes have been made and tested.

gregcaporaso:
Thanks @Clockwork-Rat! I think (@ebolyen, please correct me if I'm wrong) that we'll never have ints as input here, so it would be safe to drop your check and just always round. 

@hagenjp, do you mind pulling this down and testing the changes like you did with the last iteration? Note that in this iteration, both `summarize` and `summarize-plus` should fail if provided with a `FeatureTable[RelativeFrequency]`. 

@Clockwork-Rat, can you also add a note to the release notes document that the interface of `summarize` will be changing in 2024.2 to no longer accept a `FeatureTable[RelativeFrequency]`? We should discuss next week whether it makes sense to add another action to use in its place that generates more relevant information.

hagenjp:
@Clockwork-Rat @gregcaporaso These changes look good to me! 
I was able to see that relative frequency tables were  absent from the  accepted table formats in the --help output for feature-table summarize and summarize plus. And when I provided a relative frequency table as an input both summarize and summarize plus both returned errors for invalid value for --i-table.

q2_feature_table/_summarize/_visualizer.py
```diff
@@ -356,12 +361,43 @@ def _frequencies(table, axis):
...
+    nonzeros = table.matrix_data.data
+
+    _round = True
+
+    for item in nonzeros:
+        if not is_integer(item):
+            _round = False
+            break
+
+    first = frequencies.quantile(0.25)
```
gregcaporaso, line 367: @Clockwork-Rat, a lot of this logic seems left over from the handling of relative frequencies - is that right? I think this can mostly be stripped at this point.

gregcaporaso (changes requested):
@Clockwork-Rat, could you a take a pass through and remove everything we don't need from this PR, now that we're not concerned about supporting relative frequency values? Sorry for the trouble - this is one of those PRs that had a big change in direction once we got into it.

q2_feature_table/_summarize/_visualizer.py
Clockwork-Rat: I will remove this, @ebolyen had told me it was necessary to leave in even with dropping support for relative frequency, but I don't believe we should be checking this anymore.

Clockwork-Rat:
@hagenjp @gregcaporaso here are the changes with the removed logic.

gregcaporaso (approved):
Looks good, thanks @Clockwork-Rat!

PR #297, opened 2023-12-12
IMP: adds underscore prefix to helper method inside filter seqs tests

### LINKED ISSUE #296: IMP: helper method in `filter_seqs` should have an underscore prefix
~~**Bug Description**:
While reviewing @hagenjp's [latest PR](https://github.com/qiime2/q2-feature-table/pull/292) in feature-table, I noticed that there was an existing test without the correct method prefix, meaning this test is never actually executed when the test suite runs.~~

_**EDIT**: this was a helper method, which is why there was no 'test' prefix on the method name. To help clarify this to other developers and provide consistent syntax with the rest of our helper methods, @hagenjp has added an underscore as the method prefix, and has updated all instances of its use._

@hagenjp I am leaving this description fairly sparse because I think this will be a good opportunity for you to do a bit of sleuthing. 🤓 

Here are a couple of questions to help get you started:

1. What is the required prefix for all unit test methods?
2. How can you compare the number of tests that exist within a given test file with the number of tests being executed through `pytest`?

I decided to create a separate issue for this since this test was unrelated (and untouched) in your recent PR. Once you know where to look, this should be a fairly quick one for you to knock out. 🙂

hagenjp:
Closes [#296](https://github.com/qiime2/q2-feature-table/issues/296)

### DIFF
--- a/q2_feature_table/tests/filter/test_filter_sequences.py
+++ b/q2_feature_table/tests/filter/test_filter_sequences.py
@@ -31,8 +31,8 @@ class FilterSeqsTests(unittest.TestCase):
         md_full.index.name = 'FeatureID'
         self.md_full = qiime2.Metadata(md_full)
 
-    def filter_and_assertEqual(self, exp, md=None, exclude_ids=False,
-                               where=None):
+    def _filter_and_assertEqual(self, exp, md=None, exclude_ids=False,
+                                where=None):
         if md is None:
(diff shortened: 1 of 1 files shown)


hagenjp:
@lizgehret this is ready for your review when you have time :)

lizgehret (approved):
lgtm, thanks @hagenjp!

PR #301, opened 2023-12-22
IMP: 'subsample' action renamed to 'subsample_ids'

### LINKED ISSUE #299: rename `subsample` action to `subsample-ids`
... to avoid confusion with `rarefy` and `*-rarefaction` actions. @ebolyen and I discussed `subsample-ids` or `subsample-axis` as more descriptive action names, reflecting the fact that it's features or samples that are subsampled from the table (rather than counts, as in `rarefy`). This is in part motivated by the discussion of definitions in [this article](https://doi.org/10.1128/msphere.00355-23). 

The pending interface change will be announced in the 2023.12 Prepare notes, so we're ok to make this change for 2024.2.

hagenjp:
Closes: [#299 ](https://github.com/qiime2/q2-feature-table/issues/299)

### DIFF
--- a/q2_feature_table/tests/test_subsample.py
+++ b/q2_feature_table/tests/test_subsample.py
@@ -11,17 +11,18 @@ from unittest import TestCase, main
 import numpy as np
 import numpy.testing as npt
 from biom.table import Table
+from biom.table import subsample
 
 from q2_feature_table import subsample_ids
 
 
-class SubsampleIDsTests(TestCase):
+class SubsampleTests(TestCase):
 
     def test_subsample_samples(self):
         t = Table(np.array([[0, 1, 3], [1, 1, 2]]),
                   ['O1', 'O2'],
                   ['S1', 'S2', 'S3'])
-        a = subsample_ids(t, 2, 'sample')
+        a = subsample(t, 2, 'sample')
         self.assertEqual(a.shape, (2, 2))
 
         sample_ids = frozenset(a.ids(axis='sample'))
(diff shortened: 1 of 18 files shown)


lizgehret (changes requested):
Hey @hagenjp! This mostly looks good, but the tests are failing because you've replaced the name of an external method (`Table.subsample`) with `subsample_ids`, which doesn't exist. I'm going to let you find the instances of this and replace them with the original method, because I think it's a good exercise to recognize the difference between external methods that we're calling and our own internal methods (within QIIME 2). In your basecamp message, you already identified where that external method is being called from, so I'll let you extrapolate from there 🙂 

Let me know if you have any questions!

hagenjp:
@lizgehret this pr is ready for round two review from you :)

q2_feature_table/tests/test_subsample.py
```diff
@@ -12,7 +12,7 @@
 import numpy.testing as npt
 from biom.table import Table
 
-from q2_feature_table import subsample
+from q2_feature_table import subsample_ids
 
 
 class SubsampleTests(TestCase):
```
lizgehret:
```suggestion
class SubsampleIDsTests(TestCase):
```
I think you had the right idea when you changed the name of this class @hagenjp - it's consistent with the name changes you've implemented within all of the subsequent tests, so changing the class name as well makes sense to me!

lizgehret (approved):
lgtm, just one small comment inline!

lizgehret:
looks great, thanks @hagenjp!

PR #302, opened 2024-02-22
BUG: Summarize Index Patch

### LINKED ISSUE #303: sample ids not displayed in summarized feature table
Using the *Moving Pictures* tutorial feature table and sample metadata, the displayed per-sample counts in the `feature-table summarize` (and `summarize-plus`) visualizations doesn't have the sample ids, but rather what looks like default `pandas.Index` values. This can be reproduced as follows with those inputs: 

```
qiime feature-table summarize --i-table table.qza --m-sample-metadata-file sample-metadata.tsv --output-dir out
```

Here's a screenshot of the output:

Screenshot 2024-02-22 at 10 34 50 AM

Also note that the highlighting seems to be off (three rows are red, when four should be), and the samples are sorted by ascending counts (rather than descending counts).

Clockwork-Rat:
Fixes issue with sample IDs being lost to default indexes.

### DIFF
--- a/q2_feature_table/tests/test_summarize.py
+++ b/q2_feature_table/tests/test_summarize.py
@@ -383,6 +383,16 @@ class SummarizeTests(TestCase):
             index_fp = os.path.join(output_dir, 'index.html')
             self.assertTrue(os.path.exists(index_fp))
 
+            sample_frequency_fp = os.path.join(output_dir,
+                                               'sample-frequency-detail.html')
+            self.assertTrue(os.path.exists(sample_frequency_fp))
+
+            with open(sample_frequency_fp) as sp:
+                text = sp.read()
+                self.assertTrue('S1' in text)
+                self.assertTrue('S2' in text)
+                self.assertTrue('S3' in text)
+
     def test_frequency_ranges_are_zero(self):
         table = biom.Table(np.array([[25, 25, 25], [25, 25, 25]]),
                            ['O1', 'O2'],
--- a/q2_feature_table/tests/test_summarize.py
+++ b/q2_feature_table/tests/test_summarize.py
@@ -10,6 +10,7 @@ import os
 from unittest import TestCase, main
 import tempfile
 import re
+import json
 
 import skbio
 import biom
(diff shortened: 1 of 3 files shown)


Clockwork-Rat:
@Oddant1 here is the patch

gregcaporaso:
This issue is described in https://github.com/qiime2/q2-feature-table/issues/303.

Oddant1:
@Clockwork-Rat I tested it locally and it looks like it works. We should be able to add a unit test using [this](https://pandas.pydata.org/docs/reference/api/pandas.read_html.html) pandas method that reads tables from html into dataframes.

Not sure if this will be useful to you, but I used that method [here](https://github.com/qiime2/qiime2/pull/735/files) to create a helper to assert that all tables in a given html contained no NaNs.

lizgehret:
Okay so just for posterity (in case this happens after I'm OOO) - once a test is added here and this is merged into dev, the merge commit will need to be cherry picked onto the Release-2024.2 branch. 

cc @ebolyen

ebolyen:
Hey @Clockwork-Rat and @Oddant1, can we get the tests added soon? Worst case we could put the tests in the dev branch and just cherry-pick this one commit.

q2_feature_table/tests/test_summarize.py
```diff
@@ -383,6 +384,16 @@ def test_basic(self):
...
 
+            sample_frequency_fp = os.path.join(output_dir,
+                                               'sample-frequency-detail.html')
+            self.assertTrue(os.path.exists(sample_frequency_fp))
+
+            with open(sample_frequency_fp) as fi:
+                text = fi.read()
+                tbl_rx = re.compile(
+                    r'\{*[(S3)(S2)(S3)]*[(S3)(S2)(S3)]*[(S3)(S2)(S3)]*\}')
+                self.assertTrue(tbl_rx.search(text))
```
ebolyen: Hey @Clockwork-Rat, this approach feels a little too complicated. 

It appears that the regex is redundant as it's the same capture-group group with an optional number of repeats, repeated 3 times. This makes me feel like I'm missing something about how the regex works, which also makes it less clear what the assertion is actually verifying. 

(I assume regex.search is truthy when there is any match at all, but then why the very specific regex above?)

---

What we probably want to do is find the `<script>` tag with the right `type="application/json"`. That won't be a fun thing to do, but we can probably [tolerate a regex in this case](https://stackoverflow.com/a/1732454/579416). Once you have the contents, however that came to be, you ought to be able to load it into a pandas dataframe with some kind of `from_json` method which will make it super easy to test and work with.

ebolyen (changes requested):

q2_feature_table/tests/test_summarize.py
Clockwork-Rat: On it!

Clockwork-Rat: @Oddant1, @ebolyen updated, I used the built in ```json.loads()``` function because the ```pd.read_json()``` function did not work, but the ```loads()``` function worked fine. If necessary I could pass the results into a ```DataFrame``` constructor which should work fine.

ebolyen (approved):
Looks good, thanks @Clockwork-Rat!

PR #305, opened 2024-03-01
Sort correctly in viz by ensuring we have ints

Oddant1:


Oddant1:
@ebolyen need to patch again Monday :partying_face:

q2_feature_table/_summarize/summarize_assets/sample-frequency-detail.html
```diff
@@ -262,6 +269,11 @@ <h3>Plot Controls</h3>
     }
     updateSliderVal(val);
   }
+
+  // Parse a string that may contain commas into a base 10 integer
+  function strToInt(val) {
+    return parseInt(val.replaceAll(",", ""), 10);
```
ebolyen, line 275: The only risk here is the formatting could depend on LC_NUMERIC which will define the customary radix operator for the locale. Since this happens in Python-land there's not a convenient way to notice.

Oddant1: Can't we just also replace "."? I didn't realize this was happening automatically in Python, but I guess looking at the code again that does make sense. I suppose in theory it could always be something other than "," or "." but even that isn't a big deal. We only want numbers in there, we can just replace anything that isn't a digit with "".

PR #306, opened 2024-03-05
Feature table summarize refactor

Oddant1:
Consistently format tables


Oddant1:
Probably pending a test

EDIT: Nvm will probably merge then figure out testing

lizgehret:
Overall these changes seem reasonable (and a definite improvement to the previous implementation). Couple of nitpicks from local testing.

1. The calculated means for frequency per sample & frequency per feature have a pretty crazy number of decimal places. Not sure if we need that much precision? Maybe something like 2 decimal places is enough?

Screen Shot 2024-05-16 at 3 21 23 PM

3. Commas are no longer included in the frequency under the interactive sample detail tab (which I personally think is a visual improvement), but they are still included in the frequency under the feature detail. I don't think this is a huge issue but maybe it would be nice to remove commas from the feature detail frequencies as well (if it's easy to do):

Screen Shot 2024-05-16 at 3 22 38 PM
Screen Shot 2024-05-16 at 3 22 58 PM

lizgehret (commented):
Overall these changes seem reasonable (and a definite improvement to the previous implementation). Couple of nitpicks from local testing.

1. The calculated means for frequency per sample & frequency per feature have a pretty crazy number of decimal places. Not sure if we need that much precision? Maybe something like 2 decimal places is enough?
Screen Shot 2024-05-16 at 3 21 23 PM

2. Commas are no longer included in the frequency under the interactive sample detail tab (which I personally think is a visual improvement), but they are still included in the frequency under the feature detail. I don't think this is a huge issue but maybe it would be nice to remove commas from the feature detail frequencies as well (if it's easy to do):

Screen Shot 2024-05-16 at 3 22 38 PM
Screen Shot 2024-05-16 at 3 22 58 PM

colinvwood (approved):

cherman2 (approved):
This was just a user review from me since others have done code review. Looks good to me.

gregcaporaso:
Looks good to me too!

PR #315, opened 2024-08-01
MAINT: update split to ignore samples that are in the metadata but not the feature table

### LINKED ISSUE #314: feature-table split fails if there are samples in the metadata that are not in the feature table

gregcaporaso:
fixes #314


lizgehret (approved):
this lgtm, thanks @gregcaporaso!

PR #317, opened 2024-10-23
ENH: Added "seed" parameter to `subsample-ids`

### LINKED ISSUE #55: add option for seeding rarefaction?
**Improvement Description**
A forum user [suggested](https://forum.qiime2.org/t/seeding-the-rarefaction-algorithm/76) that we add support for seeding rarefaction, which is an interesting idea for supporting reproducibility, though I'm not certain what the specific use cases would be. 

**Questions**
Are there times where we would want to perfectly replicate rarefaction results? If so, we'd need the seed to be logged into the artifact's provenance. 

**References**
[suggested](https://forum.qiime2.org/t/seeding-the-rarefaction-algorithm/76)

### LINKED ISSUE #320: Add "seed" parameter to subsample-ids
- A seed parameter should be added to subsample-ids similar to how it is described in https://github.com/qiime2/q2-feature-table/issues/55 
- The seed will enhance reproducibility for the function.

VinzentRisch:
solves #320

- Adds seed parameter to `subsample-ids` function.
- Seed is per default set to 1. It can be set to "random" what creates a random 32 bit int as the seed. 
- Setting the seed to default 1 was done to ensure that the seed is logged in provenance. This improved reproducability.
- At this point it is not possible to have a random seed that is created in the function that will be recorded in provenance.


q2_feature_table/_subsample_ids.py
```diff
@@ -5,12 +5,14 @@
...
 # ----------------------------------------------------------------------------
+from typing import Union
 
 import biom
+import numpy as np
 
 
 def subsample_ids(table: biom.Table, subsampling_depth: int,
-                  axis: str) -> biom.Table:
+                  axis: str, seed: Union[int, str] = 1) -> biom.Table:
```
gregcaporaso:
```suggestion
                  axis: str, seed: Union[int, None] = None) -> biom.Table:
```

gregcaporaso (changes requested):
@VinzentRisch, I thought I left a comment on this earlier today, but I see to have lost it so I'm thinking maybe it didn't save and I navigated away from the page. But, apologies if this feedback is showing up twice. 

I completely agree that we need to be storing random seeds in provenance, but the approach implemented here will be problematic. The vast majority of users won't notice or override this parameter setting, in which case the result will be that the same seed is used all the time. 

We have support for passing random seeds on the command line in q2-boots - see how we handle this here:
https://github.com/caporaso-lab/q2-boots/commit/9195f9cfac1694196e1dfcc68934301d9238b350

Could you update this PR to use this approach? 

I created a framework issue to develop an approach for better supporting this:
https://github.com/qiime2/qiime2/issues/823

q2_feature_table/_subsample_ids.py
```diff
@@ -20,8 +22,14 @@ def subsample_ids(table: biom.Table, subsampling_depth: int,
                          'elements on the desired axis. The maximum depth '
                          'is: %d.' % len(table.ids()))
 
+    # Generate a random seed if seed is None
+    if seed == "random":
```
gregcaporaso: You can use the built-in [`random.seed` function](https://docs.python.org/3/library/random.html#random.seed) for this.

gregcaporaso:
Oh, I see - my previous comment was on #321. This feedback is relevant to both of the PRs.

q2_feature_table/_subsample_ids.py
gregcaporaso:
```suggestion
                  axis: str, seed: int = None) -> biom.Table:
```

q2_feature_table/_subsample_ids.py
```diff
@@ -20,8 +22,14 @@ def subsample_ids(table: biom.Table, subsampling_depth: int,
...
+        rng = np.random.default_rng()
+        seed = rng.integers(0, 2 ** 32 - 1)
+
```
gregcaporaso: This should all be able to go away since this happens in `Table.subsample` - right? 

```suggestion
```

q2_feature_table/plugin_setup.py
```diff
@@ -61,7 +61,9 @@
     function=q2_feature_table.subsample_ids,
     inputs={'table': FeatureTable[Frequency]},
     parameters={'subsampling_depth': Int % Range(1, None),
-                'axis': Str % Choices(['sample', 'feature'])},
+                'axis': Str % Choices(['sample', 'feature']),
+                'seed': Int % Range(0, 2**32) | Str % Choices(["random"])
```
gregcaporaso:
```suggestion
                'seed': Int % Range(0, None)
```
```suggestion
                'seed': Int % Range(0, 2**32) | Str % Choices(["random"])
```

q2_feature_table/plugin_setup.py
```diff
@@ -71,7 +73,11 @@
                               'the resulting table.'),
         'axis': ('The axis to sample over. If "sample" then samples will be '
                  'randomly selected to be retained. If "feature" then '
-                 'a random set of features will be selected to be retained.')
+                 'a random set of features will be selected to be retained.'),
+        'seed': ('Set the seed for the subsampling. Using the same seed with '
+                 'the same table will always lead to the same result. Using '
+                 '"random", sets the seed to a random number. The random '
+                 'seed will not be logged in provenance.')
```
gregcaporaso: Edits to the text. I'd leave the bit about provenance out, since we'll ideally be fixing that very shortly and it's common to all actions that take seeds right now. 

```suggestion
        'seed': ('Set the seed for the subsampling. Using the same seed with '
                 'the same table will always lead to the same result. By '
                 'default a random seed will be used.')
```

q2_feature_table/tests/test_subsample.py
```diff
@@ -21,7 +21,7 @@ def test_subsample_samples(self):
         t = Table(np.array([[0, 1, 3], [1, 1, 2]]),
                   ['O1', 'O2'],
                   ['S1', 'S2', 'S3'])
-        a = subsample_ids(t, 2, 'sample')
+        a = subsample_ids(t, 2, 'sample', 'random')
```
gregcaporaso: Can you adapt these tests to be similar to the ones we put together for #321?

gregcaporaso (changes requested):
@VinzentRisch, could you update this PR to mirror the approaches we've taken to this functionality and the tests in #321?

gregcaporaso:
@VinzentRisch - is this still something you're interested in working on?

VinzentRisch:
Hey @gregcaporaso 
Sorry! this got lost on my todo list. I will take a look today.

VinzentRisch:
Hey @gregcaporaso 
Again sorry for the oversight. I added all the changes similar to https://github.com/qiime2/q2-feature-table/pull/321.

gregcaporaso (approved):
@VinzentRisch, this looks good, thank you! I made a fix (and a simplification) to the tests [here](https://github.com/qiime2/q2-feature-table/pull/317/commits/939dde3930507f51478fdea486259f20346df28f). The tests were all right, but the message that would have been printed on failure (regarding the use or non-use of a random seed) was backwards such that a confusing message would have been printed on failure. 

This is ready for merge once the tests pass.

PR #321, opened 2024-12-05
ENH: Adds seed parameter to `rarefy`

### LINKED ISSUE #55: add option for seeding rarefaction?

VinzentRisch:
closes #55 
- Adds seed parameter to `subsample-ids` function.
- Seed is per default set to 1. It can be set to "random" what creates a random 32 bit int as the seed. 
- Setting the seed to default 1 was done to ensure that the seed is logged in provenance.
- At this point it is not possible to have a random seed that is created in the function that will be recorded in provenance.


gregcaporaso:
@VinzentRisch, I completely understand the motivation for this, but I don't think this solution will work out as-is. The vast majority of users wouldn't change this setting from the default, which would have the effect of there being no randomization of the seed, and repeat runs would use the same seed. 

At the moment, we don't have a way to store the seed in provenance, but we know that this is something that should be addressed. What I would recommend is that you make the default "random", and if you want to make sure your seed is stored in provenance, you can generate a random in outside of QIIME 2 and pass that in. Do you want to make that change, and then ping me for re-review? Or let me know if you'd like to discuss more.

gregcaporaso (changes requested):
See comments on this PR and on #317

VinzentRisch:
Okay that makes sense. I'll make the changes.

VinzentRisch:
Hi @gregcaporaso 
I made the changes.

q2_feature_table/_normalize.py
```diff
@@ -5,17 +5,26 @@
...
-def rarefy(table: biom.Table, sampling_depth: int,
-           with_replacement: bool = False) -> biom.Table:
+def rarefy(table: biom.Table,
+           sampling_depth: int,
+           with_replacement: bool = False,
+           random_seed: float = None
+           ) -> biom.Table:
+    # Generate a random seed if seed is None
+    if random_seed is not None:
+        random.seed(random_seed)
```
gregcaporaso: I think this should get you what you are looking for:

```suggestion
    random.seed(a=random_seed)
```

From [the docs](https://docs.python.org/3/library/random.html#random.seed):

> If a is omitted or None, the current system time is used. If randomness sources are provided by the operating system, they are used instead of the system time (see the os.urandom() function for details on availability).
> If a is an int, it is used directly.

q2_feature_table/_normalize.py
```diff
@@ -5,17 +5,26 @@
...
+
     if with_replacement:
         table = table.filter(lambda v, i, m: v.sum() >= sampling_depth,
                              inplace=False, axis='sample')
+
     table = table.subsample(sampling_depth, axis='sample', by_id=False,
-                            with_replacement=with_replacement)
+                            with_replacement=with_replacement,
+                            seed=random.randint(0, 2**32 - 1))
```
gregcaporaso: I think you don't want this, as you're overriding whatever the user provided here. Also, you will have already set the seed earlier in this function. 

```suggestion
                            with_replacement=with_replacement)
```

q2_feature_table/tests/test_normalize.py
```diff
@@ -17,7 +17,7 @@
 
 class RarefyTests(TestCase):
 
-    def test_rarefy(self):
+    def test_rarefy_random_seed(self):
```
gregcaporaso: @VinzentRisch, this test isn't actually testing that the seed is randomized. I suggest a test like the following. This can fail intermittently, but that should be rare (and if it fails too often, we can increase `n_iterations`. Note that I wrote this on GitHub, so I haven't tested that it works (there may be syntax errors/etc). 

```suggestion
    def test_rarefy_random_seed_is_randomized(self):
        t = Table(np.array([[0, 1, 3], [1, 1, 2]]),
                  ['O1', 'O2'],
                  ['S1', 'S2', 'S3'])
        a = rarefy(t, 2)
        a_eq_b = []
        n_iterations = 100
        for i in range(n_iterations):
            b = rarefy(t, 2)
            self.assertEqual(b.shape, (2, 2))
            self.assertEqual(set(b.ids(axis='sample')), set(['S2', 'S3']))
            self.assertEqual(set(b.ids(axis='observation')), set(['O1', 'O2']))
            npt.assert_array_equal(b.sum(axis='sample'), np.array([2., 2.]))
            a_eq_b.append(a == b)
        self.assertTrue(False in a_eq_b,
                                  f"After {n_iterations} iterations, all resulting feature "
                                  "tables are identical. It therefore seems that a "
                                  "randomized seed is not being used.")

    def test_rarefy(self):
```

q2_feature_table/tests/test_normalize.py
```diff
@@ -27,6 +27,14 @@ def test_rarefy(self):
...
         npt.assert_array_equal(a.sum(axis='sample'), np.array([2., 2.]))
 
+    def test_rarefy_seed_1(self):
+        t = Table(np.array([[0, 1, 3], [1, 1, 2]]),
+                  ['O1', 'O2'],
+                  ['S1', 'S2', 'S3'])
+        a = rarefy(t, 2, random_seed=1)
+        self.assertEqual(a.data('S2', axis='sample').tolist(), [1, 1])
+        self.assertEqual(a.data('S3', axis='sample').tolist(), [1, 1])
+
```
gregcaporaso: I would implement a similar test here, b/c even if a randomized seed is used you could get unlucky here and have the tables match anyway. I think this might be the case in your current implementation (see my comment above about passing a random seed through to `Table.subsample`). Alternatively, you could keep your existing `self.assertEquals` statements in this case, but I would put these in a loop over `n_iterations` to make sure you're not getting unlucky. 

```suggestion
    def test_rarefy_seed_is_not_randomized(self):
        t = Table(np.array([[0, 1, 3], [1, 1, 2]]),
                  ['O1', 'O2'],
                  ['S1', 'S2', 'S3'])
        a = rarefy(t, 2, random_seed=1)
        a_eq_b = []
        n_iterations = 100
        for i in range(n_iterations):
            b = rarefy(t, 2, random_seed=1)
            a_eq_b.append(a == b)
        self.assertFalse(False in a_eq_b,
                                  f"After {n_iterations} iterations, at least one feature "
                                  "table differed from the others. It therefore seems that a "
                                  "randomized seed is being used.")

gregcaporaso (changes requested):
@VinzentRisch, thanks for the edits. I have additional specific suggestions on this implementation.

VinzentRisch:
Hi @gregcaporaso 
I'm sorry about the last changes. But now it should be correct. You can just set the seed parameter of subsample to None, that creates a random seed internally in the function. I also implemented your tests with the iteration. 
There are some tests for the heat maps failing in the CI, I don't really know what the problem is.

gregcaporaso:
Thanks @VinzentRisch - I will review this week.

gregcaporaso (approved):
Thanks for the updates @VinzentRisch. I tested this locally and it's working for me. This is ready for merge, pending the failing heatmap tests. 

I'm not sure what's going on with the failing heatmap tests, but I don't think it's anything that you're doing. I'll discuss with the folks on my end to figure out what's going on and then get this merged.

gregcaporaso:
@VinzentRisch, do you have an option to "Update branch" on this pull request, as described [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/keeping-your-pull-request-in-sync-with-the-base-branch). If so, could you do that? I want to see if it addresses the test failures.

VinzentRisch:
I updated the branch, but the tests are still failing.

gregcaporaso:
Thanks @VinzentRisch, we'll look into it from here. I don't think it's anything that you're doing wrong.

gregcaporaso:
Alright, we got it sorted out @VinzentRisch - merged. Thanks again for the work on this one, and for your patience with the back-and-forth!

PR #331, opened 2025-03-13
MAINT: `pkg_resources` deprecation

lizgehret:


ebolyen:
looks like it was angry after all

PR #339, opened 2025-06-30
MAINT: updated verbiage in summarize viz

### LINKED ISSUE #313: Table Summary should say "Number of Unique Features"
Feature-table summarize uses  the  term "feature" in 2 seperate ways. 

In the Table Summary it is being used to discribe the unique features

Screenshot 2024-07-24 at 3 36 31 PM

In the Sample Details Page it is being use to describe all sequences. This is causing confusion. [Forum ex-ref](https://forum.qiime2.org/t/wrong-values-at-the-percentage-of-the-retained-sequences-while-looking-at-interactive-sample-detail/6698/2)

I understand that feature is specifically a broad term that can encompass both of these instants, but I do think its alittle confusing that back to back we are using them. The could cause confusing because on the first page it says that there are 770 features but then when you are choosing an even sampling depth it says I have over 100,000 features. 

If we changed the Table Summary to be say "Number of Unique Features".  We would avoid this confusion.

lizgehret:
closes #313

### DIFF
--- a/q2_feature_table/_summarize/summarize_assets/index.html
+++ b/q2_feature_table/_summarize/summarize_assets/index.html
@@ -20,11 +20,11 @@
               <td id="number-of-samples"></td>
             </tr>
             <tr>
-              <th scope="row">Number of features</th>
+              <th scope="row">Number of unique features</th>
               <td id="number-of-features"></td>
             </tr>
             <tr>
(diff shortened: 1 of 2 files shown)


cherman2 (approved):
LGTM LGTM

PR #340, opened 2025-08-04
Fixing incorrect sequence counts in tabulate-squences

### LINKED ISSUE #316: tabulate-seqs reports the number of sequences in the input, not the number of rows in the table
If `tabulate-seqs` is called with `--p-merge-method intersect` or `--p-merge-method union`, the `Sequence Count` reported at the top left (see the star in the screenshot below) is the number of sequences in the input `data` Artifact. That may be different than the number of sequences reported on in the table (i.e., the number of table rows), which can be lower (if `--p-merge-method intersect`) or higher (if `--p-merge-method union`). The sequence length stats presented are, however, computed on the total `Sequence count` that is reported. 

See these lines:
https://github.com/qiime2/q2-feature-table/blob/a4d0e2e9c3c541ea6a41a8e583564c83c2f64095/q2_feature_table/_summarize/_visualizer.py#L51
https://github.com/qiime2/q2-feature-table/blob/a4d0e2e9c3c541ea6a41a8e583564c83c2f64095/q2_feature_table/_summarize/_visualizer.py#L75
and notice that `seq_lengths` isn't filtered in the `--p-merge-method intersect` case. 

In the `--p-merge-method intersect` it would probably make the most sense to have the `Sequence Count` and length stats reflect the sequences that are reported on in the table. 

In the `--p-merge-method union` case on the other hand, it's possible that we won't have sequences for some of the table rows, in which case the sequence length stats would reflect only what's present in the input `data` Artifact, so it probably makes sense to have `Sequence Count` reflect what's in the input data. 

In both cases, we should be making it more clear specifically what `Sequence Count` is a count of. 

Screenshot 2024-10-03 at 12 09 41 PM

Macabe222:
The sequence count and sequence length stats will now reflect the data passed in for `merge-method=union`, and reflects the number of sequences kept after intersecting for `merge-method=intersect` in `tabulate-seqs`

### DIFF
--- a/q2_feature_table/tests/test_summarize.py
+++ b/q2_feature_table/tests/test_summarize.py
@@ -92,6 +92,10 @@ class TabulateSeqsTests(TestCase):
             _compute_descriptive_stats(seq_lengths)
 
     def test_seq_count_union(self):
+        """
+        This tests that the sequence count is equal to the sequence count in
+        the DNA iterator passed into to `tabulate_seqs`.
+        """
         seqs = DNAIterator(skbio.DNA(a, metadata=b)for a, b in (
             ('ACGT', {'id': 'seq01'}),
             ('AAAA', {'id': 'seq02'}),
@@ -133,10 +137,16 @@ class TabulateSeqsTests(TestCase):
 
             with open(Path(output_dir) / 'descriptive_stats.tsv', 'r') as f:
                 lines = f.readlines()
-                self.assertTrue('count\t8' in line for line in lines)
+                for line in lines:
+                    print(line)
+                self.assertTrue('count\t5\n' in lines)
 
     def test_seq_count_intersection(self):
-        seqs = DNAIterator(skbio.DNA(a, metadata=b) for a, b in (
+        """
+        This tests that the sequence count reflects the intersection of the
+        data, metadata, and taxonomy.
+        """
+        seqs = DNAIterator(skbio.DNA(seq, metadata=md) for seq, md in (
             ('AAAA', {'id': 'seq02'}),
             ('GGGG', {'id': 'seq03'}),
             ('GGGG', {'id': 'seq04'}),
@@ -176,7 +186,7 @@ class TabulateSeqsTests(TestCase):
 
             with open(Path(output_dir) / 'descriptive_stats.tsv', 'r') as f:
                 lines = f.readlines()
-                self.assertTrue('count\t3' in line for line in lines)
+                self.assertTrue('count\t3\n' in lines)
 
     def test_descriptive_stats_integration(self):
         seqs = DNAIterator(skbio.DNA(a, metadata=b)for a, b in (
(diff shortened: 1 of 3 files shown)


q2_feature_table/tests/test_summarize.py
```diff
@@ -90,6 +91,93 @@ def test_no_sequences(self):
         with self.assertRaisesRegex(ValueError, 'No values provided.'):
             _compute_descriptive_stats(seq_lengths)
 
+    def test_seq_count_union(self):
+        seqs = DNAIterator(skbio.DNA(a, metadata=b)for a, b in (
```
colinvwood, line 99:
```suggestion
        seqs = DNAIterator(skbio.DNA(seq, metadata=md) for seq, md in (
```

q2_feature_table/tests/test_summarize.py
```diff
@@ -90,6 +91,93 @@ def test_no_sequences(self):
...
+            ('ACGT', {'id': 'seq01'}),
+            ('AAAA', {'id': 'seq02'}),
+            ('GGGG', {'id': 'seq03'}),
+            ('GGGG', {'id': 'seq04'}),
+            ('TTTT', {'id': 'seq05'})
+        ))
+
+        metadata = pd.DataFrame(index=['seq01', 'seq02',
```
colinvwood, line 107: break "index", "columns", "data" down so that they're only indented once

q2_feature_table/tests/test_summarize.py
```diff
@@ -90,6 +91,93 @@ def test_no_sequences(self):
...
+                output_dir,
+                seqs,
+                taxonomy=taxonomy,
+                metadata=metadata,
+                merge_method='union'
+            )
+
+            with open(Path(output_dir) / 'descriptive_stats.tsv', 'r') as f:
+                lines = f.readlines()
+                self.assertTrue('count\t8' in line for line in lines)
```
colinvwood:
```suggestion
                self.assertTrue('count\t8\n' in lines)
```

q2_feature_table/tests/test_summarize.py
```diff
@@ -90,6 +91,93 @@ def test_no_sequences(self):
...
+                output_dir,
+                seqs,
+                taxonomy=taxonomy,
+                metadata=metadata,
+                merge_method='intersect'
+            )
+
+            with open(Path(output_dir) / 'descriptive_stats.tsv', 'r') as f:
+                lines = f.readlines()
+                self.assertTrue('count\t3' in line for line in lines)
```
colinvwood:
```suggestion
                self.assertTrue('count\t3\n' in lines)
```

q2_feature_table/tests/test_summarize.py
colinvwood: same for all the other big constructors

q2_feature_table/tests/test_summarize.py
```diff
@@ -90,6 +91,93 @@ def test_no_sequences(self):
         with self.assertRaisesRegex(ValueError, 'No values provided.'):
             _compute_descriptive_stats(seq_lengths)
 
+    def test_seq_count_union(self):
```
colinvwood, line 94: docstring quickly explaining what the test does, same for other test

colinvwood (approved):

Issue #341, opened 2025-09-22
IMP: Add warnings for importing/merging taxonomies with single/mismatched levels


lizgehret:
[forum xref](https://forum.qiime2.org/t/feature-request-warnings-for-importing-taxonomies-with-odd-levels/33664)

seems reasonable and straightforward to implement. could be a good one for @Macabe222 to take if/when he has bandwidth!

PR #343, opened 2025-10-20
Creates warnings when merging tables with different taxon depth or one level of depth. 

Macabe222:
Adds warnings for merging tables with differing levels of taxonomic depth, and adds a warning when merging tables with one one level of taxonomic depth in `merge_taxa`.

Addresses https://github.com/qiime2/q2-feature-table/issues/341

### DIFF
--- a/q2_feature_table/_merge.py
+++ b/q2_feature_table/_merge.py
@@ -71,11 +71,29 @@ def merge_seqs(data: pd.Series) -> pd.Series:
     return _merge_feature_data(data)
 
 
-def merge_taxa(data: pd.DataFrame) -> pd.DataFrame:
-    data = _merge_feature_data(data)
+def merge_taxa(data: list[pd.DataFrame]) -> pd.DataFrame:
+
+    frame_one = data[0]
+    frame_two = data[1]
+
+    count = frame_one["Taxon"].str.count(';')
+    count_two = frame_two["Taxon"].str.count(';')
+
+    if not count.equals(count_two):
+        raise ValueError(
+            "You are trying to merge tables with different levels of taxonomic"
+            " depth, this may cause failures later."
+        )
+    if 0 in count.values:
+        raise ValueError(
+            "You are trying to merge tables with one level of taxonomic"
+            " depth, this may cause failures later."
+        )
+
     # merge orders columns alphabetically; Taxon must be first header column
     # as defined here: https://github.com/qiime2/q2-types/blob/
     # 067d83e2aefe98674433e95162336fb5b9d96474/q2_types/feature_data/
     # _format.py#L97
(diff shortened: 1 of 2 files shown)


lizgehret:
Hey @Macabe222 is this ready for a review?

Macabe222:
@lizgehret Yes it is

lizgehret:
Sounds good, thanks @Macabe222! We'll get someone to look at this either this week or next 🙂

q2_feature_table/_merge.py
```diff
@@ -72,10 +73,35 @@ def merge_seqs(data: pd.Series) -> pd.Series:
 
 
 def merge_taxa(data: pd.DataFrame) -> pd.DataFrame:
-    data = _merge_feature_data(data)
+    if isinstance(data, list):
```
colinvwood, line 76: `data` is always a list because it's typed as one (see `plugin_setup.py`)

q2_feature_table/_merge.py
```diff
@@ -72,10 +73,35 @@ def merge_seqs(data: pd.Series) -> pd.Series:
...
+        frame_one = data[0]
+        frame_two = data[1]
```
colinvwood, line 78: This doesn't work for multiple reasons. First, the input is a list even if there is only one taxonomy passed in--this will give an index error for `data[1]`. Second, this doesn't handle the case where more than two taxonomies are being merged.

q2_feature_table/_merge.py
```diff
@@ -72,10 +73,35 @@ def merge_seqs(data: pd.Series) -> pd.Series:
...
+
+        count = frame_one["Taxon"].str.count(';')
+        count_two = frame_two["Taxon"].str.count(';')
+
+        for index, value in count.items():
```
colinvwood, line 83: What this does is see if the two taxonomies have the same number of levels *for each pair of taxa*, matched by index. This is meaningless because taxon index has no meaning. What the issue is requesting is to see if each of the merged taxonomies *as a whole* has the same number of levels (maximum depth) as the others.

q2_feature_table/_merge.py
```diff
@@ -72,10 +73,35 @@ def merge_seqs(data: pd.Series) -> pd.Series:
...
+                if value != count_two[index]:
+                    warnings.warn(
+                        "You are trying to merge tables with different levels "
+                        "of taxonomic depth, this may cause failures later.",
+                        UserWarning
+                    )
+            except KeyError:
+                continue
+
+        if (count == 0).any():
```
colinvwood, line 94: This is not what the issue is requesting. Go read the [forum post](https://forum.qiime2.org/t/feature-request-warnings-for-importing-taxonomies-with-odd-levels/33664) more carefully.

PR #350, opened 2026-01-19
ENH: change output of normalize action from FeatureTable[Normalized] to FeatureTable[Frequency % Properties("normalized", "method")]

### LINKED ISSUE #349: Change output of normalize action from FeatureTable[Normalized] to FeatureTable[Frequency % Properties("method")]
Currently the action normalize produces artifacts of type FeatureTable[Normalized] but this type cannot be used by any other action as input. The FeatureTable[Normalized] is the same as a FeatureTable[Frequency] just the counts are normalized by some method. 
This is why the output of the normalize action should be chnaged from FeatureTable[Normalized] to FeatureTable[Frequency % Properties("method")], where method describes the method used for normalisation.

VinzentRisch:
closes #349 

- Changes output of normalize action from FeatureTable[Normalized] to FeatureTable[Frequency % Properties("normalized", "_method_")]. where method describes the normalisation method used.

Test data:

[Archive 2.zip](https://github.com/user-attachments/files/24763300/Archive.2.zip)

### DIFF
--- a/q2_feature_table/plugin_setup.py
+++ b/q2_feature_table/plugin_setup.py
@@ -752,7 +752,7 @@ plugin.methods.register_function(
             0, 1, inclusive_start=True, inclusive_end=True
         ),
     },
-    outputs=[("normalized_table", FeatureTable[Frequency % P_normalized_table])],
+    outputs={"normalized_table": FeatureTable[Frequency % P_normalized_table]},
     input_descriptions={
         "table": "Feature table with gene counts.",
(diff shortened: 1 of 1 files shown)


ebolyen (approved):
LGTM. We'll need to note this in the changelog as a breaking change, but it won't be an interesting one because there are no consumers of FeatureTable[Normalized] yet, so this ends up vastly increasing the utility of this method.

Failures on Linux are from a stale prepare, OS X is passing.

PR #352, opened 2026-01-28
IMP: Swap rarefy random_seed to CaptureHolder

Oddant1:
Use `CaptureHolder` to track value of `random_seed` in provenance.


q2_feature_table/tests/test_normalize.py
```diff
@@ -27,11 +28,11 @@ def test_rarefy_random_seed_is_randomized(self):
         t = Table(np.array([[0, 1, 3], [1, 1, 2]]),
                   ['O1', 'O2'],
                   ['S1', 'S2', 'S3'])
-        a = rarefy(t, 2)
+        a = rarefy(t, 2, random_seed=FakeCaptureHolder())
```
ebolyen: I think the number of these needed might suggest something is off.

Should we consider that we either need to test the Plugin side of things, so that random_seed is initialized, or should we alter `rarefy` such that `None` is also fine (which is what the type technically says...). I'm leaning towards the latter, since Rachis would be obeying the signature by always providing a capture-holder, and it makes it easier to work with for the tests.

We could also imagine a test utility/decorator which injects any Rachis objects into the function, but I don't think that makes the tests any easier to write or think about.

q2_feature_table/_normalize.py
```diff
@@ -8,25 +8,31 @@
...
 
 def rarefy(table: biom.Table,
            sampling_depth: int,
            with_replacement: bool = False,
-           random_seed: int = None
+           random_seed: CaptureHolder = None
            ) -> biom.Table:
+    if random_seed.value is None:
+        random_int = random.randrange(RNG_MAX_SIZE)
+        random_seed.set_value(random_int)
```
ebolyen: Based on the above comment about the number of instances we need for a test, what if we imagined something like this:

```suggestion
    if random_seed is None or random_seed.value is None:
        random_int = random.randrange(RNG_MAX_SIZE)
    else:
        random_int = random_seed.value

    if random_seed and random_seed.value is None:
        random_seed.set_value(random_int)
```
This also ends up being kind of bad...

So what about this:


```suggestion
    random_seed = CaptureHolder.getorset(random_seed, random.randrange(RNG_MAX_SIZE))
```

Where if `random_seed` is None, it will construct one, but if it was provided an CaptureHolder, it inspects the state to determine if it was set already by the outer _bind, and if not, will instantiate itself (untethered to a provenance chain) with the secondary argument as the value.

Alternatively the result could just be the value instead of an instance, which might be easier on everybody.

```suggestion
    random_int = CaptureHolder.getorset(random_seed, random.randrange(RNG_MAX_SIZE))
```

q2_feature_table/_normalize.py
```diff
@@ -8,25 +8,31 @@
...
 
     if with_replacement:
         table = table.filter(lambda v, i, m: v.sum() >= sampling_depth,
                              inplace=False, axis='sample')
 
     table = table.subsample(sampling_depth, axis='sample', by_id=False,
                             with_replacement=with_replacement,
-                            seed=random_seed)
+                            seed=random_seed.value)
```
ebolyen:
```suggestion
                            seed=random_int)
```
only if paired with the above

q2_feature_table/tests/_util.py
```diff
@@ -0,0 +1,24 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016-2026, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+
+class FakeCaptureHolder:
```
ebolyen: I expect this would be needed in a lot of places in practice, so it would probably make sense to put it in `rachis.plugin.testing` as a Mock, but I think we might also be able to avoid it if we are a little clever.

ebolyen (changes requested):
I think we're missing a piece to make this scale across plugins effectively. I've got a few ideas below. Let me know what you think.

q2_feature_table/tests/test_normalize.py
Oddant1: Yes, I found it annoying to change them all. No, I don't think we should alter `rarefy` such that `None` is actually a valid argument to it. We want a `CaptureHolder` with `None` not an actual `None`. I don't think nuisances arising from calling the functions behind actions without the framework should be resolved by modifying the actions because we kind of fundamentally don't respect calling the function in this way in a real world context.

Oddant1:
blocked by rachis-org/rachis#921

colinvwood (approved):

PR #358, opened 2026-05-12
ENH: add core score action

VinzentRisch:
## Summary

Adds a core score action that computes per-feature prevalence-abundance scores from a relative-frequency feature table and emits them as FeatureData[Importance].

## Changes

- Adds core_score to compute scaled prevalence, scaled log mean abundance, and their product per feature.
- Registers the core_score method for FeatureTable[RelativeFrequency] inputs and FeatureData[Importance] outputs.
- Exports core_score from q2_feature_table.
- Adds targeted unit tests for default scoring, custom threshold/epsilon behavior, mean abundance over present samples, and finite scaling for constant values.

## Links

- Closes qiime2/q2-feature-table#357
- Depends on qiime2/q2-types#390

This PR was created in part with Codex 5.5

PR #360, opened 2026-05-20
ENH: add FeatureTable[RelativeFrequency] to the filter-features action

### LINKED ISSUE #359: ENH: allow FeatureTable[RelativeFrequency] for filter-features
PR #358 adds core scores, and downstream workflows need to pass relative-frequency-style values as metadata input so features can be filtered by core score.

`filter-features` currently only accepts `FeatureTable[Frequency]` and uses `min_frequency` / `max_frequency` integer thresholds. This should also support `FeatureTable[RelativeFrequency]`, where the corresponding thresholds are fractions from 0 to 1.

Proposed change:
- Allow `filter-features` to accept both `FeatureTable[Frequency]` and `FeatureTable[RelativeFrequency]`.
- Rename `min_frequency` / `max_frequency` to `min_abundance` / `max_abundance`.
- Use a TypeMap so frequency tables allow whole-number thresholds from 0 to infinity, while relative-frequency tables allow fractional thresholds from 0 to 1.
- Update the private filtering helper names to reflect abundance/fraction semantics.
- Add a regression test covering fractional filtering on a relative-frequency table.

VinzentRisch:
closes #359 

- add `FeatureTable[RelativeFrequency]` to the `filter-features` action
- `max-frequency` and `min-frequency` can only be used with `FeatureTable[Frequency]`
- `max-frequency` default was set to "inf" because of the `TypeMap` but behaviour is the same.
- the helper check_relative_frequency checks if all samples sum to 1 or 0 then it is classified as relative frequency.

This addition is needed to filter a `FeatureTable[RelativeFrequency]` with core-scores created with the new `core-score` action #358.

This PR was created in part with Codex 5.5

### DIFF
--- a/q2_feature_table/_filter.py
+++ b/q2_feature_table/_filter.py
@@ -98,14 +98,14 @@ def filter_samples(table: biom.Table, min_frequency: int = 0,
 
 
 def filter_features(table: biom.Table, min_frequency: int = 0,
-                    max_frequency: int | str = "inf", min_samples: int = 0,
+                    max_frequency: int | str = 'None', min_samples: int = 0,
                     max_samples: int = None,
                     metadata: qiime2.Metadata = None, where: str = None,
                     exclude_ids: bool = False,
                     filter_empty_samples: bool = True,
                     allow_empty_table: bool = False)\
                    -> biom.Table:
-    if max_frequency == 'inf':
+    if max_frequency == 'None':
         max_frequency = None
 
     is_relative_frequency = check_relative_frequency(table)
--- a/q2_feature_table/_filter.py
+++ b/q2_feature_table/_filter.py
@@ -91,13 +91,16 @@ def filter_samples(table: biom.Table, min_frequency: int = 0,
 
 
 def filter_features(table: biom.Table, min_frequency: int = 0,
-                    max_frequency: int = None, min_samples: int = 0,
+                    max_frequency: int | str = "inf", min_samples: int = 0,
                     max_samples: int = None,
                     metadata: qiime2.Metadata = None, where: str = None,
                     exclude_ids: bool = False,
                     filter_empty_samples: bool = True,
                     allow_empty_table: bool = False)\
                    -> biom.Table:
+    if max_frequency == 'inf':
+        max_frequency = None
+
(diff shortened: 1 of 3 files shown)


VinzentRisch:
Hey @ebolyen 

I now only allow `FeatureTable[Frequency]` to be used with the `max-frequency` and `min-frequency` parameters because for relative frequencies the filtering does not make sense in my opinion. The behaviour would be that all frequencies are summed together per feature but for relative frequencies this is not useful. 
Do you agree with this implementation?

colinvwood:
partially addresses qiime2/q2-taxa#144

q2_feature_table/_filter.py
```diff
@@ -91,20 +98,28 @@ def filter_samples(table: biom.Table, min_frequency: int = 0,
 
 
 def filter_features(table: biom.Table, min_frequency: int = 0,
-                    max_frequency: int = None, min_samples: int = 0,
+                    max_frequency: int | str = "inf", min_samples: int = 0,
```
ebolyen: It turns out q2cli will parse the default first into a set of types and comes up with `float('inf')` instead of string `"inf"`. This is unfortunate, and I really wish we had `None` working here instead.

We may need to instead just ensure that when `is_relative_frequency` we see `min_frequency == 0` and `max_frequency == None`, otherwise we error at runtime. It's not as satisfying, but I think that might just be easier for now. And we could always add it to the types later when we can talk about `None` as a first-class member of typemaps.

ebolyen (changes requested):
q2cli is being unhelpful.

VinzentRisch:
Hey @ebolyen 
What do you think of using "None" as a string? 
The problem with inf persists even when not using a TypeMap and just specifying:
'max_frequency': ((Str % Choices(['inf'])) | (Int % Range(0, None))),

The order does not change anything.

ebolyen:
If `"None"` works I am fine with that. I thought I had tested locally and ran into an issue. Maybe it was passing `"None"` explicitly on the CLI?

VinzentRisch:
"None" works for me in the python API and the CLI. In the CLI I passed None and "None" and both work.

misialq (approved):
I tested this with the CLI and it works as described 👍

PR #362, opened 2026-06-08
TEST: selenium test updates

lizgehret:
# Description

This PR does two things:
- Removes some stale pytest fixtures to skip selenium tests on ubuntu. These are no longer needed because tests aren't being run inside a container on the github runner.
- Pins selenium at 4.43.0 to workaround a bug that was introduced in 4.44.0. [issue xref](https://github.com/SeleniumHQ/selenium/issues/17448).

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -35,7 +35,7 @@ test:
   - q2templates >={{ q2templates }}
   - q2-types >={{ q2_types }}
   - pytest
-  - selenium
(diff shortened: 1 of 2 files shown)


lizgehret:
for posterity: i'm self-merging because this is mainly a version pin (and reverts a change I made back in January) and because i need this merged so that i can re-run our distro trials.

